PR #139, opened 2023-07-09
Changes in src/ resulting from the audit

IaroslavMazur:


IaroslavMazur:
@PaulRBerg, re: linting

I've tried running `forge ftm` locally, but this resulted in way more files than just the ones included in this PR to have been formatted:
image

So, I'm leaving the decision as to how to proceed with the formatting to you.

PaulRBerg:
> I've tried running forge ftm locally, but this resulted in way more files than just the ones included in this PR to have been formatted:

This is likely a Foundry configuration issue. Have you run `foundryup` recently?

What output do you get after running `pnpm lint`?

PaulRBerg:
It's also possible that `forge fmt` misbehaves on Windows - at any rate, time is precious these days so I'd rather not spend much time on fixing this formatter configuration issue. We will fix the formatting from our end.

IaroslavMazur:
> It's also possible that `forge fmt` misbehaves on Windows - at any rate, time is precious these days so I'd rather not spend much time on fixing this formatter configuration issue. We will fix the formatting from our end.

I agree, and I wouldn't like to take your time now that there are more important things for you to focus on.

Let's leave this investigation for another day, and thank you for taking care of the formatting! 🙏🏻

src/interfaces/ISablierV2ProxyTarget.sol
```diff
@@ -12,7 +12,7 @@ import { Permit2Params } from "../types/Permit2.sol";
 
 /// @title ISablierV2ProxyTarget
 /// @notice Proxy target with stateless scripts for interacting with Sablier V2, designed to be used by
-/// stream senders.
+/// stream senders, recipients and the recipients' approved operators.
```
PaulRBerg, line 15: The proxy target is designed to be used exclusively by the stream sender.

Whatever recipient-only functions (e.g., `burn`, ref https://github.com/sablier-labs/v2-periphery/pull/96) have been added only as a contingency - that is, if the proxy target ends up _accidentally_ used by stream recipients.

PaulRBerg (approved):
LGTM

PR #141, opened 2023-07-10
test: test call to "onStreamCanceled"

PaulRBerg:


IaroslavMazur (commented):
Not that it still matters, but LGTM 😆

Issue #158, opened 2023-08-01
Accept standard ERC-20 approvals in "ProxyTarget"


PaulRBerg:
## Problem

One of the biggest pain points of using Permit2 with Safe is that [all signatories need to sign the permit in a round-robin process](https://twitter.com/PaulRBerg/status/1686395627372949505).

We had a call with Utopia today to discuss a potential integration, and they said that this is one of the biggest issues with our current periphery setup.

## Solution

[Allow the periphery to be used via standard ERC-20 approvals](https://twitter.com/PaulRBerg/status/1686458086528008211).

## Proposal

We either:

1. Extend the current `ProxyTarget` to accept standard ERC-20 approvals
2. Turn the current `ProxyTarget` into an abstract with shared logic, and build two other `ProxyTargetPermit2` and `ProxyTargetERC20` to implement the Permit2 and ERC-20 variants.

PaulRBerg:
In https://github.com/sablier-labs/v2-interfaces/discussions/380#discussioncomment-6628777, @razgraf said this:

> Not to mention updates like https://github.com/sablier-labs/v2-periphery/issues/158 which are already set to make the client app more complex (through more conditional branching), thus more error prone

What does this issue (accepting standard ERC-20 approvals) have to do with the Sablier Interface?

Our official app can continue to work with Permit2 exclusively. The point is to enable more flexibility for integrators like Utopia.

razgraf:
Wouldn't our UI suffer from the same problem within the Safe integration? I was thinking we'd support both: Permit2 for cheap approvals, classic approve for async approvals.

PaulRBerg:
Aah, good point ...

I wonder if it wouldn't be easier to let users toggle the approval mechanism?

razgraf:
Yup, that's how I was thinking we should implement it. A global toggle. 

The default value could be permit2 for all deployments, for consistency, or we can make erc20-approve be the default jus for SAFEs. But that's a separate point.

PaulRBerg:
Alternatively .. make ERC-20 the default and fuggetabout Permit2 🙈

Don't see much of an improvement in UX with Permit2, TBH, even for EOAs.

PR #161, opened 2023-08-11
feat/target erc20

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/158

In the `Base_Test` I decided not to change the `defaults.permit2Params` passed in the new `data` parameter(target helper functions), as doing so would have significantly increased the complexity and I don't think is worth it for the time being. Therefore, the Permit2 struct is passed within the `data` parameter to both `targetERC20` and `targetPermit2`, but it is ignored in `targetERC20`.

@PaulRBerg, if you are not happy with this, we can discuss the solutions I tried to implement.


test/fork/plugin/onStreamCanceled.t.sol
```diff
@@ -17,6 +15,7 @@ abstract contract OnStreamCanceled_Fork_Test is Fork_Test, PermitSignature {
     function setUp() public virtual override {
         Fork_Test.setUp();
         proxyRegistry.installPlugin(plugin);
+        target = targetERC20;
```
PaulRBerg, line 18: This should be moved to `deployDeterministicPeriphery`.

src/SablierV2ProxyTargetERC20.sol
```diff
@@ -0,0 +1,31 @@
...
+pragma solidity >=0.8.19;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { SablierV2ProxyTarget } from "./abstracts/SablierV2ProxyTarget.sol";
+
+/// @title SablierV2ProxyTargetERC20
+/// @notice Proxy target contract that implements the transfer logic for standard ERC-20 approvals.
+contract SablierV2ProxyTargetERC20 is SablierV2ProxyTarget {
+    /// @notice  Parameter `data` is ignored in this implementation.
```
PaulRBerg, line 11: This is absolutely not a `@notice` information. `@notice` describes the high-level behavior of the function. Here, `@dev` should have been used.

src/SablierV2ProxyTargetERC20.sol
```diff
@@ -0,0 +1,31 @@
...
+    /// @dev See the documentation for the user-facing functions that call this internal function.
```
PaulRBerg, line 12: Given how important this function has become, we should add some proper documentation in here.

src/abstracts/SablierV2ProxyTarget.sol
```diff
@@ -208,7 +192,7 @@ contract SablierV2ProxyTarget is
         // Calculate the sum of all of stream amounts. It is safe to use unchecked addition because one of the create
         // transactions will revert if there is overflow.
         uint256 i;
-        uint160 transferAmount;
+        uint128 transferAmount;
```
PaulRBerg, line 195: This should have remained `uint160`

https://github.com/cantinasec/review-sablier2/issues/6

src/SablierV2ProxyTargetERC20.sol
```diff
@@ -0,0 +1,31 @@
...
+        bytes calldata /* data */
+    )
+        internal
+        override
+    {
+        // Retrieve the proxy owner.
+        address owner = _getOwner();
+
+        // Transfer funds from the proxy owner to the proxy.
+        asset.transferFrom({ from: owner, to: address(this), amount: amount });
```
PaulRBerg, line 26: This should use OpenZeppelin's `safeTransferFrom`.

src/abstracts/SablierV2ProxyTarget.sol
```diff
@@ -707,26 +691,14 @@ contract SablierV2ProxyTarget is
         }
     }
 
-    /// @dev Helper function to transfer funds from the proxy owner to the proxy using Permit2 and, if needed, approve
-    /// the Sablier contract to spend funds from the proxy.
+    /// @dev Helper function to transfer funds from the proxy owner to the proxy, if needed, approve the Sablier
+    /// contract to spend funds from the proxy.
```
PaulRBerg, line 695: Needs to be rewritten to explain that this function is meant to be implemented by child contracts.

PaulRBerg (changes requested):
Great work, @andreivladbrg. A couple of issues and suggestions:

First, the integration tests pass the `permit2Params` as data to both targets. This is fine from a technical point of view because `ProxyTargetERC20` ignores the data, but it is bad test design. We want to test the new target in the way it is intended to be used, i.e., with an empty value for the data parameter. To achieve this, we need to modify all `batchCreate` functions in `Base_Test`.

Second, the fork tests should be refactored to follow the `shared` directory structure implemented in `integration`. The current approach of testing both targets is inconsistent - the fork tests use different function names, and the integration tests use different contracts.

Third, the name of the newly introduced `data` parameter is vague and it also overlaps with the calldata passed to `execute` and `fallback` in PRBProxy. We should give it a more specific name such as `transferData`.

Fourth, the name of the `ProxyTargetERC20` contract is okay, but I don't like it that much because: (i) not all chains use the term ERC-20, for instance BSC uses BEP20 and (ii) only a small surface of ERC-20 is required, i.e. `transferFrom`. Here are some alternatives:

- `ProxyTargetApprove`
- `ProxyTargetApproval`
- `ProxyTargetTransfer`
- `ProxyTargetTransferFrom`
- `ProxyTargetStandard`

andreivladbrg:
> This is fine from a technical point of view because ProxyTargetERC20 ignores the data

Yeah

> To achieve this, we need to modify all batchCreate functions in Base_Test. I disagree that doing so would lead to a "significantly increased complexity".

I might not be exact with the wording, let's put this way: I couldn't find an elegant way, without adding a lot of new lines of code.

> Second, the fork tests should be refactored to follow the shared directory structure implemented in integration 

I thought about this, but I decided not to do so because we don't do this in the core either.

> the fork tests use different function names, and the integration tests use different contracts.

Why do say this? The have the same name, only they have added the "ERC20" or "Permit2".

>Third, the name of the newly introduced data parameter is vague and it also overlaps with the calldata passed to execute and fallback in PRBProxy. We should give it a more specific name such as transferData

Ok, we can rename it to `transferData`, but the argument about the "overlaps" with parameters from other repos (which, yes, is used in our v2-periphery, but is still external) is a weak point.

> Fourth, the name of the ProxyTargetERC20 contract is okay, but I don't like it that much because:

I just followed the suggestion from the issue:
image

PaulRBerg:
@andreivladbrg I have pushed a few PRS to address all issues and comments mentioned above. PR ready for re-review.

PaulRBerg:
> I couldn't find an elegant way, without adding a lot of new lines of code.

That's fine - I think I have found a way, see `getTransferData` in `Base_Test`.

> I decided not to do so because we don't do this in the core either

In V2 Core, we don't need shared tests in the fork tests.

> Why do say this? The have the same name, only they have added the "ERC20" or "Permit2".

See what I did

>  the argument about the "overlaps" with parameters from other repos ... is a weak point.

PRBProxy is directly called by V2 Periphery. Who cares that it is an external repo?

The `data` parameters need to be prepared in V2 Periphery. So this is not a weak point, not at all.

> I just followed the suggestion from the issue:

That was just an initial conjecture - I have improved upon it with the ratioale above.

src/abstracts/SablierV2ProxyTarget.sol
andreivladbrg: Forgot about it

src/SablierV2ProxyTargetERC20.sol
andreivladbrg: Correct

script/DeployProtocol.s.sol
```diff
@@ -41,7 +41,7 @@ contract DeployProtocol is BaseScript {
         // Deploy V2 Periphery.
         archive = new SablierV2Archive(initialAdmin);
         plugin = new SablierV2ProxyPlugin(archive);
-        target = new SablierV2ProxyTarget(permit2);
+        target = new SablierV2ProxyTargetPermit2(permit2);
```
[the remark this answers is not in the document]
andreivladbrg: Yes we should

src/SablierV2ProxyTargetERC20.sol
andreivladbrg: Correct

andreivladbrg:
>  PR ready for re-review.
Looks good. I also pushed 3 more commits, please approve the PR if it is ready to be merged

shell/prepare-artifacts.sh
```diff
@@ -27,8 +27,6 @@ FOUNDRY_PROFILE=optimized forge build
 # Copy the production artifacts
 cp out-optimized/SablierV2Archive.sol/SablierV2Archive.json $artifacts
 cp out-optimized/SablierV2ProxyPlugin.sol/SablierV2ProxyPlugin.json $artifacts
-cp out-optimized/SablierV2ProxyTargetApprove.sol/SablierV2ProxyTargetApprove.json $artifacts
-cp out-optimized/SablierV2ProxyTargetPermit2.sol/SablierV2ProxyTargetPermit2.json $artifacts
```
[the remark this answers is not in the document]
PaulRBerg: Why not? The `_transferAndApprove` function is not external. The bytecodes for these contracts should be identical to `ISablierV2ProxyTarget`.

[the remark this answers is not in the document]
PaulRBerg: Buuuut it's not just about ABIs .. external users might be interested in knowing the bytecodes, too.

you are right, @andreivladbrg.

PaulRBerg:
good job @andreivladbrg.

PR #164, opened 2023-08-17
Update remappings to use "src"

### LINKED ISSUE #163: Update "@prb/proxy" and "@sablier/v2-core" import paths to use "src"
Install the latest `@sablier/v2-core`, and then update the import paths to use `src/`.

Refs:

- https://github.com/sablier-labs/v2-core/issues/646
- https://github.com/sablier-labs/v2-core/pull/648

andreivladbrg:
Closes #163


PaulRBerg (approved):
LGTM, except that we also need to update `@sablier/v2-core` in `package.json`. I pushed a commit.

PR #165, opened 2023-08-17
docs: roll 1.0.3

andreivladbrg:


PaulRBerg (approved):
You forgot to bump the `version` in `package.json`, but otherwise LGTM

PR #173, opened 2023-08-24
feat: add "ProxyTargetPush"

### LINKED ISSUE #172: Move to a push-based model in the standard ERC20 proxy target
The idea is:

1. Have the user make a standard transfer to the proxy
2. Call the updated proxy target
3. Expect that the proxy has sufficient tokens to cover the `totalAmount`, otherwise, let the ERC-20 contract revert the transaction due to insufficient funding.

Doing so would have two benefits:

1. It would let us abstract away the proxy deployment by calling `deployAndExecuteInstallPlugin` in the Sablier UI (because with a push-based model, there is no need to call `_getOwner()`)
2. It would mitigate the ERC-20 allowance issues reported by @tibbarytsur in https://github.com/cantinasec/review-pr-sablier/issues/1

PaulRBerg:
Closes #172.


andreivladbrg (changes requested):
In the `expectCallToTransfer` and `expectCallToTransferFrom` from `Base_Test` we have to add:

```solidity
if (target != targetPush) {}
```

We should add the `SablierV2ProxyTargetPush` in the `prepare-artifacts` script. Add a workflow dispatch.

Regarding the naming, I don't think "Push" is the best name, it doesn't really say much. What about "NoTransfer"

PaulRBerg:
Thanks for the review.

> In the `expectCallToTransfer` and `expectCallToTransferFrom` from `Base_Test` we have to add:

Good catch.

> We should add the `SablierV2ProxyTargetPush` in the `prepare-artifacts` script. Add a workflow dispatch.

Yes.

> Regarding the naming, I don't think "Push" is the best name

It's a good name, actually.

1. [This is how people typically refer to `transfer`-based flows](https://twitter.com/0xM3rlin/status/1686522018785820672).
2. It's a short, succinct name (just one word.

andreivladbrg:
>It's a good name, actually.
    [This is how people typically refer to transfer-based flows](https://twitter.com/0xM3rlin/status/1686522018785820672).


Ok, fine. It is the first time I hear the word in this context, and maybe that's why it doesn't sounds well with me.

Also, there is one more miss, we should remove `SafeERC20` from the contract: 

https://github.com/sablier-labs/v2-periphery/blob/af807da8d3f634ea72191be29bd23d3fff272dbb/src/SablierV2ProxyTargetPush.sol#L12

andreivladbrg (approved):

Issue #168, opened 2023-08-22
Contract for batch creating streams without proxy


PaulRBerg:
## Problem

As discussed [here](https://github.com/sablier-labs/v2-interfaces/discussions/380#discussioncomment-6768799), it is not possible to abstract away the proxy deployment in the UI because the proxy owner is an immutable variable in PRBProxy:

https://github.com/PaulRBerg/prb-proxy/blob/84ae9671d64023c0c21316392e11e33a2eaa2511/src/PRBProxy.sol#L27

Immutable variables don't exist during construction, so any attempt to read the owner when using `deployAndExecute` will revert.

## Solution

Implement a contract that can batch-create streams without relying on the proxy.

andreivladbrg:
Given that we fixed the proxy abstraction with `SablierV2ProxyTargetPush`, can we close this issue? @PaulRBerg

andreivladbrg:
Ok, should it be a push system or a classic pull?

PaulRBerg:
It has to be a classic pull because the batch contract would be the same contract for all users. The push model works only when the user is in control of the contract (as is the case with PRBProxies).

The license: `GPL-3.0-or-later`.

andreivladbrg:
> The push model works only when the user is in control of the contract

Yeah, wanted to make sure in case I was missing something out.

andreivladbrg:
I currently did the following implementation: https://github.com/sablier-labs/v2-periphery/tree/feat/batch-create

I feel the naming might be a bit repetitive(contract + functions). Especially when:

https://github.com/sablier-labs/v2-periphery/blob/ec58ee142ed2650e063e76b5fee89b7d865fdee9/test/integration/batch-create/batch-create-with-durations/batchCreateWithDurations.t.sol#L41

I was thinking of renaming all function in `SablierV2BatchCreate` like this but I think it would be misleading:
  - `batchCreateWithDeltas` -> `createWithDeltas` 
  - `batchCreateWithDurations` -> `createWithDurations` 
  -  `batchCreateWithMilestones` -> `createWithMilestones`  
  - `batchCreateWithRange` -> `batchCreateWithRange`
  
@PaulRBerg What do you think? Should we keep it as it is, or do you have better suggestions?

PaulRBerg:
`SablierV2Batch` is cool.

And removing the `batch` prefix from all those functions is cool, too.

PR #177, opened 2023-08-31
feat: add SablierV2Batch

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/168


PaulRBerg (dismissed):
Generally LGTM. I made just one small commit to adjust a few things.

There seems to be a lot of duplication between the `ProxyTarget` and the `Batch` tests. I wonder whether there's any way to reduce it?

andreivladbrg:
>There seems to be a lot of duplication between the ProxyTarget and the Batch tests. I wonder whether there's any way to reduce it?

Not sure yet, given they are different system, one is called directly(Batch) one is used to via bytes through Proxy (only delegate call) I don't think it would be best choice to dry fy the tests here.

Let's take care of this task in another branch. I've already created a separated one which is meant to reduce the `Base_Test` size: https://github.com/sablier-labs/v2-periphery/tree/test/move-target

In the mean time, can you approve? Let's not merge it yet because we have to first ship new packages to NPM for the new targets.

PaulRBerg:
>  I don't think it would be best choice to dry fy the tests here
>
> Let's take care of this task in another branch.

Makes sense

> Let's not merge it yet because we have to first ship new packages to NPM for the new targets.

The `main` branch is already "polluted" since it contains the airstreams contract - what we can do is ship a new npm package in the context of commit https://github.com/sablier-labs/v2-periphery/tree/7ba65defde936ca63de220e15d2dd9995c85547e (i.e. tag this commit with `v1.1.0` or smth). I can take care of it.

Issue #167, opened 2023-08-21
Update branching trees to use "given" keyword


PaulRBerg:
References:

- https://github.com/sablier-labs/v2-core/issues/621
- https://github.com/sablier-labs/v2-core/discussions/647
- https://github.com/sablier-labs/v2-core/pull/642

smol-ninja:
Can you assign this to me? I can work on this.

andreivladbrg:
Thanks for the interest, I just finished working on this branch: https://github.com/sablier-labs/v2-periphery/tree/test/improve-tree-wording. Was about to create a PR.

PR #178, opened 2023-09-01
test: use "given" keyword in branching trees

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/167

### DIFF
--- a/test/integration/airstream/campaign-factory/create-airstream-campaign-ll/createAirstreamCampaignLL.t.sol
+++ b/test/integration/airstream/campaign-factory/create-airstream-campaign-ll/createAirstreamCampaignLL.t.sol
@@ -13,7 +13,7 @@ contract CreateAirstreamCampaignLL_Integration_Test is Airstream_Integration_Tes
     }
 
     /// @dev This test works because a default campaign is deployed in {Integration_Test.setUp}
-    function test_CreateAirstreamCampaignLL_AlreadyDeployed() external {
+    function test_RevertGiven_AlreadyDeployed() external {
         bytes32 merkleRoot = defaults.merkleRoot();
         uint40 expiration = defaults.EXPIRATION();
         bool cancelable = defaults.CANCELABLE();
@@ -37,11 +37,11 @@ contract CreateAirstreamCampaignLL_Integration_Test is Airstream_Integration_Tes
         });
     }
 
-    modifier notAlreadyDeployed() {
+    modifier givenNotAlreadyDeployed() {
         _;
     }
 
-    function testFuzz_CreateAirstreamCampaignLL(address admin, uint40 expiration) external notAlreadyDeployed {
+    function testFuzz_CreateAirstreamCampaignLL(address admin, uint40 expiration) external givenNotAlreadyDeployed {
         vm.assume(admin != users.admin.addr);
         address expectedCampaignLL = computeCampaignLLAddress(admin, expiration);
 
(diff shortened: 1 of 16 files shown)


PaulRBerg (approved):
LGTM except for some hiccups, corrected in my commit.

andreivladbrg:
I see that you changed the tree branches to "given" for the `expiration` variable:

https://github.com/sablier-labs/v2-periphery/blob/d2935b860154e46c64d7b86d28f9536360f69432/test/integration/airstream/campaign-ll/clawback/clawback.tree#L5-L7

You've mentioned before that for time variables we should use "when", why is it not the case here?

Also, you changed: `given the claim is not included in the Merkle tree` to `when the claim is not included in the Merkle tree`, I'm not sure I agree with this. We store the store the `merkleRoot` which indicates whether a claim is included or not in the Merkle tree.

PaulRBerg:
> You've mentioned before that for time variables we should use "when", why is it not the case here?

There's no time variable provided as a function parameter here. There are only two variables:

- `expiration` (which is a fixed background property)
- `block.timestamp` (which is also a background property)

> We store the store the `merkleRoot` which indicates whether a claim is included or not in the Merkle tree.

This distinction between "given" and "when" is fairly fuzzy, but my rationale here was thus:

The Merkle root is a single, fixed `bytes32` variable. There's nothing that it is *prepared* when the user specifies a valid or an invalid claim. The latter is a user-provided function parameter, which the user has control over.

andreivladbrg:
> There's no time variable provided as a function parameter here. There are only two variables:

I was not talking about function parameters. See these examples:

https://github.com/sablier-labs/v2-core/blob/d7ea52b333dbb1c16af2e827653f141977491cb5/test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree#L3-L7

https://github.com/sablier-labs/v2-core/blob/d7ea52b333dbb1c16af2e827653f141977491cb5/test/integration/concrete/lockup/withdraw/withdraw.tree#L35-L39

We are talking about the time variables that are stored within `_streams` mapping and not a function parameters. It's not clear to me why is `expiration` different here.

The changes were made by you on this [commit](https://github.com/sablier-labs/v2-core/pull/656/commits/aedff61e39fa0652d314dda43b498a0b4488c8a2#diff-2dd17740b516660565534e98ea0ffe97b79f5a7fa970430602ed4cf322ca1b13).


> The Merkle root is a single, fixed bytes32 variable. There's nothing that it is prepared when the user specifies a valid or an invalid claim. 

Whether is a single bytes32 variable or not, `merkleRoot` is the only state variable within our contracts that indicates if a claim is _included_ or not in the Merkle tree. 

>The latter is a user-provided function parameter, which the user has control over.

Yes, this is why we still use "when" here:

https://github.com/sablier-labs/v2-periphery/blob/20e25c6a153ed1b8261fa6acbfcd27ee95d930d5/test/integration/airstream/campaign-ll/claim/claim.tree#L9-L15

I highly suggest to switch back to "given" for _included_ branch.

PaulRBerg:
> I was not talking about function parameters. See these examples:

Ah, good catch. Those should be updated to use "given".

> We are talking about the time variables that are stored within _streams mapping and not a function parameters. It's not clear to me why is expiration different here.

I may have made a mistake in my previous explanations.

The time variables should be considered "when" only when they are passed as function parameters (e.g. in the "create" functions).

> The changes were made by you on this [commit](https://github.com/sablier-labs/v2-core/pull/656/commits/aedff61e39fa0652d314dda43b498a0b4488c8a2#diff-2dd17740b516660565534e98ea0ffe97b79f5a7fa970430602ed4cf322ca1b13).

Yep, the changes in `streamedAmountOf` are mistaken.

> `merkleRoot` is the only state variable within our contracts that indicates if a claim is included or not in the Merkle tree.

Good point. Let's use "given" for this.

> I highly suggest to switch back to "given" for included branch.

Ok. Can you implement these changes yourself?

andreivladbrg:
> Can you implement these changes yourself?

Addressed in my latest commit, can you approve the PR?

Issue #44, opened 2023-05-11
Upgrade to latest Solidity


PaulRBerg:
Same as https://github.com/sablierhq/v2-core/issues/483.

PR #187, opened 2023-09-15
build: change solidity version to 0.8.20

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/44


PaulRBerg (changes requested):
We don't need to bump the pragma to `>=0.8.21`. We can keep it to `>=0.8.19` to enable compatibility with as many external projects as possible.

The reason why we have to use `0.8.19` is because this is what is used by PRBMath (which uses user-defined operators).

andreivladbrg:
Pushed a commit to switch back to `>=0.8.19`. 

Note: I've added a struct for the return values in the `DeployDeterministicPeriphery` script because we were facing the stack too deep error https://github.com/sablier-labs/v2-periphery/actions/runs/6196363653/job/16822859098.

Lmk if there is anything else. If not, can you approve the PR?

andreivladbrg:
@PaulRBerg Pushed a commit  to fix the conflicts, does it look good to you?

PaulRBerg (approved):
Yep, looks good now. Good job!

Issue #185, opened 2023-09-11
Incorrect type returned by "getMerkleStreamers"


PaulRBerg:
The type returned by `getMerkleStreamers` should be `ISablierV2MerkleStreamer ` instead of `ISablierV2MerkleStreamer LL`:

https://github.com/sablier-labs/v2-periphery/blob/f8a14edb11cac2e0467f543d0a706dcc7d0010f6/src/SablierV2MerkleStreamerFactory.sol#L26-L34

PaulRBerg:
@andreivladbrg could you pick up this task, please?

andreivladbrg:
Yep 👍

andreivladbrg:
Shouldn't we also change the `_merkleStreamers` mapping ?

https://github.com/sablier-labs/v2-periphery/blob/f8a14edb11cac2e0467f543d0a706dcc7d0010f6/src/SablierV2MerkleStreamerFactory.sol#L20

PaulRBerg:
Yep, we should.

PR #190, opened 2023-09-20
refactor: change the interface stored in `_merkleStreamers`

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/185 and addresses feedback left here https://github.com/sablier-labs/v2-periphery/discussions/189


src/interfaces/ISablierV2MerkleStreamerFactory.sol
```diff
@@ -52,7 +53,7 @@ interface ISablierV2MerkleStreamerFactory {
     /// @param streamDurations The durations for each stream due to the recipient.
     /// @param cancelable Indicates if each stream will be cancelable.
     /// @param ipfsCID Metadata parameter emitted for indexing purposes.
-    /// @param campaignTotalAmount Total amount of ERC-20 assets to be streamed.
+    /// @param aggregateAmount Total amount of ERC-20 assets to be streamed by all recipients.
```
PaulRBerg, line 56: "to all recipients"

PaulRBerg (approved):
LGTM, let's just fix the comment below.

I like the "aggregate amount" wording a lot. What would you say about using it in the batch create functions, too? That is, instead of `totalTransferAmount`, we would say `aggregateAmount`.

andreivladbrg:
> What would you say about using it in the batch create functions, too? That is, instead of totalTransferAmount, we would say aggregateAmount.

Both names convey the idea that the variable is a sum of multiple amounts, but I believe that `totalTransferAmount` is more specific and appropriate, it describes what the variable's purpose is in a straightforward manner, the amount is going to be transferred. However, `aggregateAmount` is more general and could be used in a wider range of contexts, like the one we use in the factory contract, only to emit in the event. It doesn't specify what the aggregate amount is for. 

Having said that, I suggest we keep `totalTransferAmount` in the batch create functions. Do you agree?

PaulRBerg:
> `totalTransferAmount` is more specific and appropriate

You may have just gotten used to it. It's not that "specific" and "appropriate" given that it can easily be confounded for each stream's `totalAmount` (notice that the term "total" can be found in both names).

> the amount is going to be transferred

Between the user and the proxy, or between the proxy and Sablier?

> It doesn't specify what the aggregate amount is for.

We will define it in the NatSpec as the sum of all per-stream `totalAmount`s.

> I suggest we keep totalTransferAmount in the batch create functions. Do you agree?

No. But we can debate this disagreement somewhere else. Happy to merge this PR now.

PaulRBerg (approved):
looking good now

PR #191, opened 2023-09-28
build: bump core

andreivladbrg:
Update the `v2-core` branch to main and add the `transferable` parameter from this [PR](https://github.com/sablier-labs/v2-core/pull/668).

Notes:
1. I have not changed yet the `package.json` file for the new `v2-core` version because it's not released on npm. Installing it as a github package would not work due to [prepack](https://github.com/sablier-labs/v2-core/blob/e626e6f9f9c460a8b42b3501de71debb684965b3/package.json#L60) from core. It would fail as seen [here](https://app.warp.dev/block/dfG3EBwso2os6NbedCqQuh).
2. At the moment we are deploying the core contracts in the fork tests because using the old version of them would not work due to the new boolean added.


src/abstracts/SablierV2MerkleStreamer.sol
```diff
@@ -35,6 +35,9 @@ abstract contract SablierV2MerkleStreamer is
     /// @inheritdoc ISablierV2MerkleStreamer
     bytes32 public immutable override merkleRoot;
 
+    /// @inheritdoc ISablierV2MerkleStreamer
+    bool public immutable override transferable;
```
PaulRBerg, line 39: We need to define this before `merkleRoot` to save gas by tight packing.

PaulRBerg (changes requested):
looks good except for that gas optimization

src/abstracts/SablierV2MerkleStreamer.sol
PaulRBerg: Update: it is not true that storage variables are tightly packed like struct fields.

- https://chat.openai.com/share/67619484-f686-4d11-8cac-cb9e8d4ca1ce
- https://ethereum.stackexchange.com/q/155029/24693

PaulRBerg (approved):
lgtm

PaulRBerg:
great job, @andreivladbrg!!

src/abstracts/SablierV2MerkleStreamer.sol
smol-ninja: ChatGPT seems to have given an incorrect response here. Storage variables are also tightly packed.

It's true that because of dealing with reduced-size variables in the same slot, changing the value of `y` will consume more gas (~100 in comparison) in the case of contract B, but it will also save gas (~16,000 in comparison) when setting it the first time. It depends on the use case and how many times each variable is read or written.

However, in the above code, it does not matter because immutable variables are part of the creation code and not of the contract storage.

- https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html
- https://docs.soliditylang.org/en/latest/contracts.html#immutable

Issue #197, opened 2023-10-02
Install Permit2 from Uniswap


PaulRBerg:
Uniswap has finally merged my PR:

https://github.com/Uniswap/permit2/pull/231

This means we can now install Permit2 from their GitHub repository.

PR #198, opened 2023-10-02
build: install permit2 from uniswap repo

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/197

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Common Changelog](https://common-changelog.org).
 - **Breaking**: Change `permit2Params` to `transferData`, which has type `bytes`
   ([#161](https://github.com/sablier-labs/v2-periphery/pull/161))
 - Bump V2 Core ([#191](https://github.com/sablier-labs/v2-periphery/pull/191))
+- Install `permit2` depencency from Uniswap repository ([#198](https://github.com/sablier-labs/v2-periphery/pull/198))
 - Upgrade Solidity to `0.8.21` ([#187](https://github.com/sablier-labs/v2-periphery/pull/187))
 
 ### Added
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,9 @@
   path = "lib/openzeppelin-contracts"
   url = "https://github.com/OpenZeppelin/openzeppelin-contracts"
 [submodule "lib/permit2"]
-  branch = "release"
+  branch = "main"
   path = "lib/permit2"
-  url = "https://github.com/sablier-labs/permit2"
+  url = "https://github.com/uniswap/permit2"
 [submodule "lib/prb-proxy"]
   branch = "release-v4"
   path = "lib/prb-proxy"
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,10 +14,11 @@ The format is based on [Common Changelog](https://common-changelog.org).
 
 ### Changed
 
-- **Breaking**: Change `permit2Params` to `transferData`, which has type `bytes`
-  ([#161](https://github.com/sablier-labs/v2-periphery/pull/161))
+- Install `permit2` depencency from Uniswap repository ([#198](https://github.com/sablier-labs/v2-periphery/pull/198))
 - Bump V2 Core ([#191](https://github.com/sablier-labs/v2-periphery/pull/191))
 - Upgrade Solidity to `0.8.21` ([#187](https://github.com/sablier-labs/v2-periphery/pull/187))
+- **Breaking**: Change `permit2Params` to `transferData`, which has type `bytes`
```
PaulRBerg, line 20: Breaking updates need to be at the top

CHANGELOG.md
```diff
@@ -14,10 +14,11 @@ The format is based on [Common Changelog](https://common-changelog.org).
 
 ### Changed
 
-- **Breaking**: Change `permit2Params` to `transferData`, which has type `bytes`
-  ([#161](https://github.com/sablier-labs/v2-periphery/pull/161))
+- Install `permit2` depencency from Uniswap repository ([#198](https://github.com/sablier-labs/v2-periphery/pull/198))
```
PaulRBerg, line 17: Not ordered alphabetically

PaulRBerg (changes requested):
thanks for the quick turnaround 😅

We also have to update the Permit2 entry in `.gitmodules`

CHANGELOG.md
andreivladbrg: Why does it have to alphabetically? For it makes more sense to order chronologically

andreivladbrg: The common changelog standard doesn't require alphabetical order

PaulRBerg: Not all entries will have associated PRs so it will be difficult to remember the order in which the changes had been implemented

andreivladbrg: chronologically, i.e. the order the PRs were created, in this case: 
- 198
- 191
- 187
- 161

andreivladbrg: But I agree **Breaking** should be at top

PaulRBerg: @andreivladbrg we have PRs for each change, but this may not always be the case. See for example the [PRBMath changelog](https://github.com/PaulRBerg/prb-math/blob/main/CHANGELOG.md).

Can we please use alphabetic ordering?

andreivladbrg:
>We also have to update the Permit2 entry in .gitmodules

Yep, realised just after I created the PR

Pushed a new commit, does it look good now?

PaulRBerg (approved):
looks good

Issue #199, opened 2023-10-02
Capitalize immutable variables


PaulRBerg:
Immutable variables are constant at runtime, so they should be named like constants in `UPPERCASE_SNAKE_CASE`.

Solhint has a rule for this, which is currently disabled in our Solhint config file. Activating it finds several occurrences:

https://app.warp.dev/block/SeqReMKq5UCYkpi0rjOZuO

We should refactor and then activate the rule (by removing [this line](https://github.com/sablier-labs/v2-periphery/blob/c1acdfe9dd63e903a639de04727b64ffacd7da22/.solhint.json#L12) from the Solhint config).

PR #200, opened 2023-10-02
refactor: capitalize immutable variables

andreivladbrg:
Depends on https://github.com/sablier-labs/v2-periphery/pull/198

Closes https://github.com/sablier-labs/v2-periphery/issues/199

Note: being a small change I think this shouldn't be included in the `CHANGELOG` .


PaulRBerg:
@andreivladbrg can you rebase from `main` please

In the future, consider branching off from `main` (`git checkout -b NEW_BRANCH main`) so that PRs don't depend on each other.

andreivladbrg:
But my intention was to depend on that PR 😅

This way I avoided potential conflicts.

PaulRBerg:
> This way I avoided potential conflicts.

There were no _actual_ conflicts between these two PRs. And this can often be inferred between two small PRs.

PaulRBerg (changes requested):
LGTM except for the `_original` immutable var in `OnlyDelegateCall`, which also has to be modified.

andreivladbrg:
> There were no actual conflicts between these two PRs. And this can often be inferred between two small PRs.

Yes there would have been conflicts, in the `.gitmodules` file. A small one, but still a conflict. It does not matter the size of a PR, for me is a good heurstic to prevent any potential conflict.

> LGTM except for the _original immutable var in OnlyDelegateCall, which also has to be modified.

Correct, pushed a commit now

PaulRBerg:
> for me is a good heurstic to prevent any potential conflict.

Fair enough

PaulRBerg (approved):
looks good now.

We will have to announce this PR in the changelog because the ABI has changed in the `ProxyPlugin` - `archive` is now `ARCHIVE`.

Issue #201, opened 2023-10-05
Add Merkle root in "CreateMerkleStreamerLL" event


PaulRBerg:
As per @razgraf's feedback provided on Slack, this event is missing the Merkle root:

https://github.com/sablier-labs/v2-periphery/blob/c970d08c746545d3bff000834b5e897f2fac8a14/src/interfaces/ISablierV2MerkleStreamerFactory.sol#L19C1-L31

PR #203, opened 2023-10-06
perf: emit merkle root in createMerkleStreamerLL function

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/201


PaulRBerg (approved):
Good work

Issue #204, opened 2023-10-06
Adapt leaf hashing algorithm for MerkleStreamer


razgraf:
In the MerkleStreamer campaign contract we use the following algorithm to construct our leaf:

1. we tightly encode three values (index, address, amount)
2. we create a keccak hash over them

https://github.com/sablier-labs/v2-periphery/blob/4b27ec08637514719b21a7780832bfd0d41b1c84/src/SablierV2MerkleStreamerLL.sol#L78

This is consistent with Uniswap's [approach](https://github.com/Uniswap/merkle-distributor/blob/25a79e8ec8c22076a735b1a675b961c8184e7931/contracts/MerkleDistributor.sol#L47), which unfortunately does not look like standard OpenZeppelin as far as leaf construction goes. 

As we had to replicate this algorithm outside of the contract (Javascript and Rust) we've bumped into some issues related to this choice:
1. `abi.encodePacked` is [harder](https://github.com/ethers-io/ethers.js/issues/718) (and less safe) to reproduce than `abi.encode`
2. the usage of a single keccak seems to go against security practices such as the ones against [second preimage attacks](https://flawed.net.nz/2018/02/21/attacking-merkle-trees-with-a-second-preimage-attack/)
3. all OpenZeppelin utilities, including the [JS library](https://github.com/OpenZeppelin/merkle-tree/#validating-a-proof-in-solidity) and the [Rust crate](https://lib.rs/crates/merkle-tree-rs) use `encode` and 2x `keccak` which makes compatibility not only a good argument for us but for external integrators as well 

---

@andreivladbrg @gavriliumircea and I are currently testing a version of our Merkle Streamer periphery with the following structure for generating leafs:
```solidity
bytes32 leaf = keccak256(bytes.concat(keccak256(abi.encode(index, recipient, amount))));
```

If successful, we propose swapping this algorithm in the `main` branch, inside the `claim` method.

PaulRBerg:
Great idea and write-up. It seems like the right approach to switch to `abi.encode` and hash twice instead of just once.

Issue #205, opened 2023-10-09
Use Sorted Data Leaves in `MerklerStreamer` tests


andreivladbrg:
We are currently testing the `MerklerStreamer` contract using a tree whose hashed data leaves are not in ascending order: 

https://github.com/sablier-labs/v2-periphery/blob/db29e2034a99a2ffc665fcb14b2bfe516ffe51a4/test/utils/Defaults.sol#L73-L79

This isn’t ideal for our purposes. In a production environment, the Merkle tree will contain data leaves that are sorted in ascending order.

I suggest we implement an algorithm for quick sorting an array of `bytes32`

PR #206, opened 2023-10-09
refactor: double hash

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/204 and https://github.com/sablier-labs/v2-periphery/issues/205


PaulRBerg (changes requested):
I have stopped my review when I saw the `binarySearch` and `sort` functions implemented in our code base.

Both OpenZeppelin and Solady offer implementations for this:

- [`findUpperBound`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5161a4dc8a9475343290f4811d3e725c49fab1bb/contracts/utils/Arrays.sol#L24)
- [`searchSorted`](https://github.com/Vectorized/solady/blob/08901b1131be4903e989884571198d72de1c0ee9/src/utils/LibMap.sol#L204)

There's no point in maintaining mirror implementations of the same code, so we should just import and use one of these two libraries

andreivladbrg:
Firstly, there is no function for `sort` implemented by either OpenZeppelin or solady, please review carefully what that function does. Secondly, both versions you mentioned use a `uint256` parameter, not `bytes32`, tho this isn't ultimately the big problem because conversion is possible. The main issue with these functions is that they require a `storage` parameter, not a `memory` or `calldata` parameter, which is what we need.

Was getting this error:
image

PaulRBerg:
There are mupltiple `sort` functions available in Solady - see [`LibSort`](https://github.com/Vectorized/solady/blob/08901b1131be4903e989884571198d72de1c0ee9/src/utils/LibSort.sol).

Then, it is true that the [`searchSorted`](https://github.com/Vectorized/solady/blob/08901b1131be4903e989884571198d72de1c0ee9/src/utils/LibMap.sol) function require a `storage` variable, but why is that a problem?

You can just define a storage array in the `Defaults` contract, like this:

```solidity
bytes32[] public leaves = new bytes32[](RECIPIENTS_COUNT);

constructor() {
    leaves[0] = MerkleBuilder.computeLeaf(INDEX1, users.recipient1.addr, CLAIM_AMOUNT);
    leaves[1] = MerkleBuilder.computeLeaf(INDEX2, users.recipient2.addr, CLAIM_AMOUNT);
    leaves[2] = MerkleBuilder.computeLeaf(INDEX3, users.recipient3.addr, CLAIM_AMOUNT);
    leaves[3] = MerkleBuilder.computeLeaf(INDEX4, users.recipient4.addr, CLAIM_AMOUNT);
}
```

In general, I would like us to prefer to import code rather than copy-paste it, so that we don't have to (i) understand it and (ii) maintain it.

PaulRBerg:
Perfect, thanks!

PaulRBerg (approved):
Generally LGTM, let's just add the explanatory comment mentioned below

andreivladbrg:
I see that you want to remove the `_initMerkleTree` function, I believe it would be better to keep it because we would have all the logic inside the `MERKLE STREAMER` header "scope" so that when you read those constants/variables you don't have to switch back and forth to constructor

PaulRBerg:
I insist on keeping those declarations in the `constructor`, because they pertain to initializing immutable variables. See my latest commit.

I've also noticed that the constructor was defined in the middle of the contract, and so I've moved it to the top.

I think it's looking good now!

andreivladbrg:
> I insist on keeping those declarations in the constructor, because they pertain to initializing immutable variables

well, in case of `LEAVES` arrays cannot be constant/immutable

PaulRBerg (approved):
looks good

Issue #208, opened 2023-10-12
Find a way to reduce the number of lines of code in the "Base_Test" contract


PaulRBerg:
The [`Base_Test`](https://github.com/sablier-labs/v2-periphery/blob/db29e2034a99a2ffc665fcb14b2bfe516ffe51a4/test/Base.t.sol) contract is clogged up; it's got 569 lines of code at the time of posting this issue.

It might be worth exploring the possibility of moving some of the utils in a separate abstract contract, similar to [how we did in V2 Core](https://github.com/sablier-labs/v2-core/blob/8ca98d3ffad2dd75e0a705ffc0cd5e798acc8058/test/utils/Utils.sol).

PR #209, opened 2023-10-12
Reduce number of lines of code in the "Base_Test" contract

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/208

### DIFF
--- a/test/Base.t.sol
+++ b/test/Base.t.sol
@@ -94,6 +94,19 @@ abstract contract Base_Test is Assertions, Events, Merkle, StdCheats, V2CoreUtil
                                      HELPERS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @dev Approves relevant contracts to spend assets from some users.
+    function approveContracts() internal {
+        // Approve Permit2 to spend assets from the stream's recipient.
+        vm.startPrank({ msgSender: users.recipient0.addr });
+        asset.approve({ spender: address(permit2), amount: MAX_UINT256 });
+
+        // Approve Permit2, SablierV2Batch and Alice's Proxy to spend assets from Alice (the proxy owner).
+        changePrank({ msgSender: users.alice.addr });
+        asset.approve({ spender: address(batch), amount: MAX_UINT256 });
+        asset.approve({ spender: address(permit2), amount: MAX_UINT256 });
+        asset.approve({ spender: address(aliceProxy), amount: MAX_UINT256 });
+    }
+
     /// @dev Generates a user, labels its address, and funds it with ETH.
     function createUser(string memory name) internal returns (Account memory user) {
         user = makeAccount(name);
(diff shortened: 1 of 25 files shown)


test/Base.t.sol
```diff
@@ -35,12 +34,13 @@ import { SablierV2ProxyTargetPush } from "src/SablierV2ProxyTargetPush.sol";
...
 import { Assertions } from "./utils/Assertions.sol";
 import { Defaults } from "./utils/Defaults.sol";
+import { DeployPrecompiled } from "./utils/DeployPrecompiled.sol";
 import { Events } from "./utils/Events.sol";
 import { Merkle } from "./utils/Murky.sol";
 import { Users } from "./utils/Types.sol";
 
 /// @notice Base test contract with common logic needed by all tests.
-abstract contract Base_Test is Assertions, Events, Merkle, StdCheats, V2CoreUtils {
+abstract contract Base_Test is Assertions, DeployPrecompiled, Events, Merkle, V2CoreUtils {
```
PaulRBerg, line 43: I would prefer to still inherit from `StdCheats` here, for maximum clarity.

PaulRBerg, line 43: I would prefer to still inherit from StdCheats here, for maximum clarity.

test/Base.t.sol
```diff
@@ -64,9 +64,9 @@ abstract contract Base_Test is Assertions, Events, Merkle, StdCheats, V2CoreUtil
     ISablierV2ProxyPlugin internal plugin;
     IPRBProxyRegistry internal proxyRegistry;
     ISablierV2ProxyTarget internal target;
-    SablierV2ProxyTargetApprove internal targetApprove;
-    SablierV2ProxyTargetPermit2 internal targetPermit2;
-    SablierV2ProxyTargetPush internal targetPush;
+    ISablierV2ProxyTarget internal targetApprove;
+    ISablierV2ProxyTarget internal targetPermit2;
+    ISablierV2ProxyTarget internal targetPush;
```
PaulRBerg, line 69: Why change the types here?

PaulRBerg (changes requested):
I skimmed through the PR and it generally looks good to me, except for one thing.

The name you have given to `DeployPrecompiled` is confusing, since the `Precompiles` contract contains similar deployment functions.

A much clearer name would be `DeployOutOptimized` (Or just `DeployOptimized`) to signal the fact that that helper contract deploys the contract artifacts found in `out-optimized`.

test/Base.t.sol
andreivladbrg: tbh I don't know why 😅

test/Base.t.sol
andreivladbrg: We can't inherit in both contracts(DeployOptimized, Base_Test), we would get the "Linearization of inheritance graph impossible" error

andreivladbrg:
I pushed a commit to address your comments.

> A much clearer name would be DeployOutOptimized (Or just DeployOptimized) to signal the fact that that helper contract deploys the contract artifacts found in out-optimized.

I went for `DeployOptimized`

LMK if there is anything else, if not can you approve it?

PaulRBerg:
Thanks!

> LMK if there is anything else

yes. Let's also rename the functions within `DeployOptimized` to `deployOptimized`, e.g. `deployOptimizedBatch`.

And we should also do this in V2 Core.

andreivladbrg:
> Let's also rename the functions within DeployOptimized to deployOptimized, e.g. deployOptimizedBatch.

Just pushed a commit

Issue #212, opened 2023-10-18
Remove "ProxyPlugin" and "Archive"


PaulRBerg:
Remove "ProxyPlugin" and "Archive".

Consequence of https://github.com/sablier-labs/v2-core/pull/710.

andreivladbrg:
Also `SablierV2Archive`

PR #213, opened 2023-10-18
Remove "ProxyPlugin" and "Archive"

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/212

For historical reasons, I have pushed a [branch](https://github.com/sablier-labs/v2-periphery/tree/plugin-archive) to keep the contracts, in case we need them at some point in the future.


PaulRBerg:
> For historical reasons, I have pushed a [branch](https://github.com/sablier-labs/v2-periphery/tree/plugin-archive) to keep the contracts

You can remove it. Git itself is a history tracker, and we also have version tags, which we can easily checkout if the need arises for looking up these contracts in the future.

test/Base.t.sol
```diff
@@ -114,19 +108,16 @@ abstract contract Base_Test is Assertions, DeployOptimized, Events, Merkle, V2Co
         deal({ token: address(asset), to: user.addr, give: 1_000_000e18 });
     }
 
-    /// @dev Conditionally deploy V2 Periphery normally or from a source precompiled with `--via-ir`.
+    /// @dev Conditionally deploy V2 Periphery normally or from a optimized source compiled with `--via-ir`.
```
PaulRBerg, line 111: "an optimized"

PaulRBerg (approved):
lgtm

PaulRBerg:
Feel free to re-review and then we can merge, @andreivladbrg.

andreivladbrg:
> Feel free to re-review and then we can merge

I am not in favor of removing the struct because we may add more contracts in the future in this repo and we might need to add it again, I suggest we keep.

Also, I don't know why but you pushed the old v2-core version, I pushed [a commit](https://github.com/sablier-labs/v2-periphery/pull/213/commits/e1aebc23ddd12900ebb6591c3a7abb2378ac6ee6) bumping the version.

andreivladbrg:
For simplicity I am going to remove your commit, and just fix the typos

PaulRBerg:
> I am not in favor of removing the struct because we may add more contracts in the future in this repo and we might need to add it agai

I disagree on the basis of YAGNI but I'm not argue about this, feel free to do as you wish

> I don't know why but you pushed the old v2-core version

Sorry, my bad. forgot to reinit the submodules

Issue #215, opened 2023-10-25
Write a variant of "withdrawMultiple" dedicated to senders


PaulRBerg:
See the context in https://github.com/sablier-labs/v2-periphery/discussions/179.

The task is to implement a variant of `withdrawMultiple` that:

1. Takes no `to` parameter
2. Queries the `recipient` for each stream id, and pass that as `to`

PaulRBerg:
This issue has become outdated in light of https://github.com/sablier-labs/v2-periphery/issues/224.

Closing.

PR #216, opened 2023-10-25
feat: withdraw multiple to recipient

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/215


src/abstracts/SablierV2ProxyTarget.sol
```diff
@@ -167,6 +167,39 @@ abstract contract SablierV2ProxyTarget is
         lockup.withdrawMultiple(streamIds, to, amounts);
     }
 
+    /// @inheritdoc ISablierV2ProxyTarget
+    function withdrawMultipleToRecipient(
```
PaulRBerg, line 171: Can we just keep naming this `withdrawMultiple`? Overloading functions is fine.

src/interfaces/ISablierV2ProxyTarget.sol
```diff
@@ -87,6 +87,21 @@ interface ISablierV2ProxyTarget {
     )
         external;
 
+    /// @notice Withdraws assets from streams to the recipient.
```
PaulRBerg, line 90: Let's rewrite to:

> Withdraws assets to the recipient of each stream.

src/interfaces/ISablierV2ProxyTarget.sol
```diff
@@ -87,6 +87,21 @@ interface ISablierV2ProxyTarget {
...
+    ///
+    /// @dev Notes:
+    /// - Retrieves the recipient of each stream from {ISablierV2Lockup.getRecipient}.
```
PaulRBerg, line 93: `using` instead of `from`

src/abstracts/SablierV2ProxyTarget.sol
PaulRBerg, line 171: Can we just keep naming this `withdrawMultiple`? Overloading functions is fine.

test/integration/target/withdraw-multiple-to-recipient/withdrawMultipleToRecipient.t.sol
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.19 <0.9.0;
+
+import { ISablierV2Lockup } from "@sablier/v2-core/src/interfaces/ISablierV2Lockup.sol";
+
+import { Errors } from "src/libraries/Errors.sol";
+
+import { Target_Integration_Test } from "../Target.t.sol";
+
+abstract contract WithdrawMultipleToRecipient_Integration_Test is Target_Integration_Test {
```
PaulRBerg, line 10: Let's use the plural form in this name, i.e., `WithdrawMultipleToRecipients`.

PaulRBerg (changes requested):
feedback below

src/interfaces/ISablierV2ProxyTarget.sol
andreivladbrg: I followed the format used in `ISablierV2Lockup`. 

https://github.com/sablier-labs/v2-core/blob/91887daff5872c7fa83fe3644b6f7582e9bffc0d/src/interfaces/ISablierV2Lockup.sol#L294

I suggest we keep like this for consistency. Or we should change both.

src/abstracts/SablierV2ProxyTarget.sol
andreivladbrg: It wouldn't be fine, because someone integrating a target wouldn't be able to use:

```solidity
abi.encodeWithSelector(ISablierV2ProxyTarget.withdrawMultiple.selector, streamIds, to, amounts);
abi.encodeCall(ISablierV2ProxyTarget.withdrawMultiple, (lockup, streamIds, to, amounts))
```

Because it will get this error:

image

So we should have different names.

test/integration/target/withdraw-multiple-to-recipient/withdrawMultipleToRecipient.t.sol
andreivladbrg: Don't understand why, can you explain?

src/interfaces/ISablierV2ProxyTarget.sol
PaulRBerg: "Formatting" is about white spaces, print widths, and other stylistic things. "Wording" is what we are talking about there, and wording does not have to be "consistent" (i.e., identical) when the context requires a different formulation.

"to the recipient of each stream" is much, much, much clearer than "from streams to the recipient" (the latter sounds like you're withdrawing to one account only).

andreivladbrg: > Formatting" is about white spaces, print widths, and other stylistic things. 

lol😅  I was not referring to this. let's not debate words, but the context was "same style or structure of phrasing" or "wording format"

> much, much clearer than "from streams to the recipient" (the latter sounds like you're withdrawing to one account only).

It is clearer. It might be me but I have this thing about "general" consistency in everything, like maker's vocabular.

Anyway, I pushed a commit with the version suggested.

PaulRBerg: Thanks!

> the context was "same style or structure of phrasing" or "wording format"

Ok, now I get it. This is not the typical way "format" is used, though.

>  like maker's vocabular.

I am staunchly against Maker's approach.

src/abstracts/SablierV2ProxyTarget.sol
PaulRBerg: Ok fine. But in this case let's use plural "recipients" instead of "recipient".

test/integration/target/withdraw-multiple-to-recipient/withdrawMultipleToRecipient.t.sol
PaulRBerg: The sender's streams could be associated with multiple, different recipients.

PaulRBerg:
Thanks Andrei.

Unfortunately, this issue has become outdated in light of https://github.com/sablier-labs/v2-periphery/issues/224.

Closing.

PR #217, opened 2023-11-20
Hash all parameters in the salt 

andreivladbrg:
We are using CREATE2 to deploy the airstream campaigns and we are not hashing all the inputs. 

This is an issue because a bad actor could match all the parameters intended by an organization, but set `streamDurations = (0, 0)`, which will unlock all tokens instantly, turning the airstream into an airdrop.

### DIFF
--- a/src/SablierV2MerkleStreamerFactory.sol
+++ b/src/SablierV2MerkleStreamerFactory.sol
@@ -56,7 +56,18 @@ contract SablierV2MerkleStreamerFactory is ISablierV2MerkleStreamerFactory {
         returns (ISablierV2MerkleStreamerLL merkleStreamerLL)
     {
         // Hash the parameters to generate a salt.
-        bytes32 salt = keccak256(abi.encodePacked(initialAdmin, lockupLinear, asset, merkleRoot, expiration));
+        bytes32 salt = keccak256(
+            abi.encodePacked(
+                initialAdmin,
+                lockupLinear,
+                asset,
+                merkleRoot,
+                expiration,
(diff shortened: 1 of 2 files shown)


PaulRBerg:
This PR addresses the following audit finding:

https://cantina.xyz/ai/6d6a47ff-5279-4502-8705-d837b159a31e/findings/a1d4e8a9-4868-49c4-a68c-db5fd6e0ba5d

src/SablierV2MerkleStreamerFactory.sol
```diff
@@ -56,7 +56,18 @@ contract SablierV2MerkleStreamerFactory is ISablierV2MerkleStreamerFactory {
...
         // Hash the parameters to generate a salt.
-        bytes32 salt = keccak256(abi.encodePacked(initialAdmin, lockupLinear, asset, merkleRoot, expiration));
+        bytes32 salt = keccak256(
+            abi.encodePacked(
+                initialAdmin,
+                lockupLinear,
+                asset,
+                merkleRoot,
+                expiration,
+                keccak256(abi.encode(streamDurations)),
```
PaulRBerg: There's no need to hash the `streamDurations`. ABI-encoding it is enough.

PaulRBerg (approved):

src/SablierV2MerkleStreamerFactory.sol
PaulRBerg: Pretty sure this is correct but I will let you @andreivladbrg confirm before squashing & merging.

PR #218, opened 2023-11-20
Remove `_merkleStreamers` mapping

andreivladbrg:
The Merkle streamer contract inherits [`Adminable`](https://github.com/sablier-labs/v2-core/blob/main/src/abstracts/Adminable.sol) contract, which allows the admin to transfer its ownership to someone else. If that happens, the `_merkleStreamers` will include incorrect data about the campaign's admin. To prevent this, I removed both the mapping and the getter function.


PaulRBerg:
Thanks @andreivladbrg, reviewing now. For future PRs, it would be helpful to mention the audit finding this PR addresses:

https://cantina.xyz/ai/6d6a47ff-5279-4502-8705-d837b159a31e/findings/6daf4dc9-65fb-4bde-87a8-bf7f42c02b30

PaulRBerg (approved):
LGTM, thanks

PaulRBerg:
Cc @razgraf @gavriliumircea, this PR entails some changes in the subgraph.

razgraf:
> Cc @razgraf @gavriliumircea, this PR entails some changes in the subgraph.

Could you go a bit into detail as to which changes may affect the subgraph? 

I don't see anything altering event signatures. Aside from maybe updating the ABIs to remove that getter (which we aren't using) is there something else we should consider?

PaulRBerg:
> Could you go a bit into detail as to which changes may affect the subgraph?

Sorry, meant to say that you should not obtain the addresses from the subgraph, not from the contract. Bad wording on my part.

PR #219, opened 2023-11-20
Add protocol fee variable in `claim` and `clawback` functions

andreivladbrg:
Note: this scenario was not going to happen, as it would have been super bad to our business.

The idea behind this is to add one more layer of security for users in case the protocol fee is changed after an airstream campaign is created. 

This would have been an issue if the expiration were set to 0, resulting in all funds being permanently locked and unable to be clawed back. Effectively, this would have allowed us to take a percentage of any airstream campaign that is created.


andreivladbrg:
Closing this PR because I used a wrong branch name

PR #220, opened 2023-11-20
 Add protocol fee variable in claim and clawback functions 

andreivladbrg:
The idea behind this is to add one more layer of security for users in case the protocol fee is changed after an airstream campaign is created. 

This would have been an issue if the protocol fee is greater than 0 and there is no expiration, resulting in funds being  locked and unable to be clawed back. Effectively, this would have allowed us to take a percentage of any airstream campaign that is created.

This PR fixes the following issue found by Zach Obront in his recent audit:

https://cantina.xyz/ai/6d6a47ff-5279-4502-8705-d837b159a31e/findings/e2ed29f3-34f2-4dc7-8849-38aa7f0c3b91


src/abstracts/SablierV2MerkleStreamer.sol
```diff
@@ -107,6 +118,15 @@ abstract contract SablierV2MerkleStreamer is
...
     function _checkClaim(uint256 index, bytes32 leaf, bytes32[] calldata merkleProof) internal view {
+        // Safe Interactions: query the protocol fee. This is safe because it's a known Sablier contract that does
+        // not call other unknown contracts.
+        UD60x18 protocolFee = LOCKUP.comptroller().protocolFees(ASSET);
+
+        // Checks: the protocol fee is zero.
+        if (protocolFee.gt(ud(0))) {
+            revert Errors.SablierV2MerkleStreamer_ProtocolFeeNotZero();
+        }
+
```
PaulRBerg, line 129: It would be better to perform the expiration check first.

- There will be many many more expired campaigns that there will be non-zero protocol fees, so making the check sooner saves gas.
- As a rule of thumb, Checks should come before Interactions (even when they are safe).

src/libraries/Errors.sol
```diff
@@ -21,18 +21,24 @@ library Errors {
...
 
-    /// @notice Thrown when trying to claim when Merkle streamer has not expired.
+    /// @notice Thrown when trying to clawback when the campaign has not expired.
     error SablierV2MerkleStreamer_CampaignNotExpired(uint256 currentTime, uint40 expiration);
 
     /// @notice Thrown when trying to claim with an invalid Merkle proof.
     error SablierV2MerkleStreamer_InvalidProof();
 
+    /// @notice Thrown when trying to clawback when the protocol fee is zero.
+    error SablierV2MerkleStreamer_ProtocolFeeZero();
```
PaulRBerg, line 34: Unused, should be deleted

src/abstracts/SablierV2MerkleStreamer.sol
```diff
@@ -86,8 +93,12 @@ abstract contract SablierV2MerkleStreamer is
 
     /// @inheritdoc ISablierV2MerkleStreamer
     function clawback(address to, uint128 amount) external override onlyAdmin {
-        // Checks: the campaign has expired.
-        if (!hasExpired()) {
+        // Safe Interactions: query the protocol fee. This is safe because it's a known Sablier contract that does
+        // not call other unknown contracts.
+        UD60x18 protocolFee = LOCKUP.comptroller().protocolFees(ASSET);
+
+        // Checks: if the protocol fee is not greater than zero, the campaign must be expired.
```
PaulRBerg, line 100: This is not a good explanation for an `AND` statement.

Let's rewrite to:

> the campaign is not expired and the protocol fee is zero

src/abstracts/SablierV2MerkleStreamer.sol
```diff
@@ -86,8 +93,12 @@ abstract contract SablierV2MerkleStreamer is
...
+        if (!protocolFee.gt(ud(0)) && !hasExpired()) {
             revert Errors.SablierV2MerkleStreamer_CampaignNotExpired({
                 currentTime: block.timestamp,
                 expiration: EXPIRATION
```
PaulRBerg, line 104: We need a new error for the case when the protocol fee is not zero.

PaulRBerg (changes requested):
See my review in the comments.

The only thing left to do is to update the `clawback` function to revert with a new custom error for the case when the protocol is non-zero.

src/abstracts/SablierV2MerkleStreamer.sol
andreivladbrg: > It would be better to perform the expiration check first.

nope, the protocol fee check has the highest priority. if there is a campaign with no expiration, this check is unnecessary, but for the protocol fee, it is always a concern if it is non-zero.

> There will be many many more expired campaigns that there will be non-zero protocol fees, so making the check sooner saves gas.

this would affect the gas only if the transaction is going to revert,  should we focus on this specific case (which is not going to happen through our UI), or on the logical aspect? it makes a lot more sense to use the protocol fee as a wall at the entrance of the function

> As a rule of thumb, Checks should come before Interactions (even when they are safe)

ya, agree, but how is it related here since we need to perform an interaction in order to make the check? no matter what you do CEI is not respected here (no grey area)

andreivladbrg:
> The only thing left to do is to update the clawback function to revert with a new custom error for the case when the protocol is non-zero.

To revert when the protocol fee is non-zero? you sure? i say that it shouldn't revert. there are 4 cases:
- `protocolFee == 0 && hasExpired == false ` --> revert
- `protocolFee == 0 && hasExpired == true` --> can clawback
- `protocolFee > 0 && hasExpired == false` --> can clawback
- `protocolFee > 0 && hasExpired == true` --> can clawback

if the protocol fee is non-zero it means that claims are unavailable, and if the expiration is set to 0 how would someone claw back the funds if we are going to revert😅?

src/abstracts/SablierV2MerkleStreamer.sol
andreivladbrg: as said below, we shouldn't

src/abstracts/SablierV2MerkleStreamer.sol
PaulRBerg: > the protocol fee check has the highest priority
> for the protocol fee, it is always a concern if it is non-zero

I don't see why it has a higher priority. They are both reverts. What matters is what revert is _more likely_ to occur.

> this would affect the gas only if the transaction is going to revert, should we focus on this specific case (which is not going to happen through our UI), or on the logical aspect?

Yes, we should focus on reverts in this case, because we should expect there to be quite a few reverted txs (not due to the protocol fee check):

1. We may make errors in how we generate the Merkle proofs
2. Other airdrop-claiming UIs (such as Earnifi) might make errors

As I've said in my previous reply, it is more likely that one of the other reverts will be triggered than the protocol fee.

>  how is this related here since we need to perform an interaction in order to make the check? no matter what you do CEI is not respected here

Making the external call after _some other checks_ is better than making it before _all checks_.

While writing this response, I realized that it would be better to move the protocol fee check at the end of the `_checkClaim` function.

PaulRBerg:
> To revert when the protocol fee is non-zero? you sure?
> if the protocol fee is non-zero it means that claims are unavailable, and if the expiration is set to 0 how would someone claw back the funds if we are going to revert😅?

LOL! Thanks for the correction.

> am i missing smth?

Nah, I have missed something. Thanks.

src/abstracts/SablierV2MerkleStreamer.sol
PaulRBerg: I pushed a commit let me know what you think @andreivladbrg

andreivladbrg: > I pushed a commit let me know what you think

i still don't agree with you here, the main reason being that it is a check that is not in control of either the claim caller or the campaign admin - this is why i say it has the higher priority.

but being a small thing, i am ok with keeping it as you want

andreivladbrg:
updated the precompiles, is this ready to be merged? @PaulRBerg

Issue #224, opened 2023-11-26
Nuke the proxy targets


PaulRBerg:
The PR that implements this issue should target the `2.2` branch because the proxy targets are already audited and we should keep them in version control just in case we need to deploy them.

- [ ] Remove all proxy targets
- [ ] Remove all scripts associated with proxy targets
- [ ] Remove all tests associated with proxy targets

andreivladbrg:
> proxy targets are already audited and we should keep them in version control just in case we need to deploy them.

hmm, don't think we should keep them. we can always use git history (checkout to a specific commit) if there's an actual need to deploy them.

the problem with keeping them in the repo (on the `main` branch) is that we won't have documentation for them, and based on the feedback received from users, we should have them synced (docs and main).

also, i think we both agree that we should not release the npm package with them, right?

i suggest we remove them for `2.1` to avert questions about them. questions like "why do you have them?" "the ui doesn't say we use a proxy," "why aren't they documented?" etc. answering a high volume of questions like these would be more time-consuming than checking out to a specific commit to deploy them, if we ever need them (but we don't know if we will ever need them again).

PaulRBerg:
good points, let's remove them

razgraf:
If the intention is for integrators to use the latest Periphery v2.1 contracts natively (no proxy), removing the targets should be the way to go. 

One important thing here is keeping historical references to v2.0 inside out documentation, given:

1. the app will still support management of v2.0 streams (e.g. the Allowance Target)
2. integrators may have already worked with v2.0 so they should be able to find the targets and ensure backwards compatibility  


I'm mentioning this for posterity as I believe this final caveat is already addressed by @andreivladbrg's idea around a nested deployment-addresses documentation. At the same time, we could move the pages addressing v2.0 targets into a "deprecated" section, just so we can keep linking to them (e.g. offer a short explainer/changelog between versions in the docs and add references to those pages).

andreivladbrg:
Addressed in 6362f2292d3d9edf642fa91ac5018958dc4e4893

PR #226, opened 2023-11-29
Remove proxy targets

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/224

Question, should we update the 2023-10-19 changelog to 2023-11-19?  or should we just updated the changes and keep the 2023-10-19 date?


src/libraries/Errors.sol
PaulRBerg, line 1: Can you please check that this does not introduce any bytecode changes between this PR and the commit that the auditors have signed off on?

If there are bytecode changes, we will keep the error.

foundry.toml
```diff
@@ -73,4 +72,4 @@
   mainnet = "${RPC_URL_MAINNET}"
   optimism = "https://optimism-mainnet.infura.io/v3/${API_KEY_INFURA}"
   polygon = "https://polygon-mainnet.infura.io/v3/${API_KEY_INFURA}"
-  sepolia = "https://sepolia.infura.io/v3/${API_KEY_INFURA}"
+  sepolia = "RPC_URL_SEPOLIA"
```
PaulRBerg, line 75: Should be `"${RPC_URL_SEPOLIA}"`

PaulRBerg (approved):
generally looks good to me, great job @andreivladbrg!

lol how much code was removed.

> or should we just updated the changes and keep the 2023-10-19 date?

In the changelog, we will put the date when we ship the npm packages. Speaking of that - when should we do it?

src/libraries/Errors.sol
andreivladbrg: this PR doesn't change the bytecode. however, it will change when we update the git submodules to NPM packages, due to different remappings

also, after the Turing audit, the bytecodes have already changed in both `v2-core` and `v2-periphery`:
- https://github.com/sablier-labs/v2-periphery/pull/227
- https://github.com/sablier-labs/v2-core/pull/735

therefore, this is not a concern.

fyi: we've also had different bytecodes after Spearbit's audit back in the summer

andreivladbrg:
> lol how much code was removed.

yep, the proxy system was quite complex 😅

> when should we do it?

we first need to publish `v2-core` so that we can install it as NPM package here

once we merge https://github.com/sablier-labs/v2-core/pull/734  i can pursue with deployments and publish

to answer your question, we will try ASAP

src/libraries/Errors.sol
PaulRBerg: Thanks for checking and for explaining.

> this is not a concern.

It is a concern because we need to ask Cantina to update the commits in the audit reports so that we can then link them on Etherscan.

The fact that we introduced some changes because of another auditor is great (it means we have an excuse for changing the bytecode).

PaulRBerg:
the PR is ready to be merged now - I will let you do it @andreivladbrg, but please make sure you squash the commits.

I suppose that the next step is to install the Node.js version of V2 Core, and subsequently remove the node modules?

PR #248, opened 2023-12-29
refactor: reorder struct fields

PaulRBerg:
Consistent ordering of the struct fields.


src/types/DataTypes.sol
```diff
@@ -54,7 +54,7 @@ library Batch {
         uint128 totalAmount;
         bool cancelable;
         bool transferable;
-        LockupLinear.Range range;
         Broker broker;
+        LockupLinear.Range range;
```
smol-ninja, line 58: Why'd you move this below `broker`? Doesn't it break consistency?

PaulRBerg: Oh, this is a mistake. Good catch.

smol-ninja (approved):

PaulRBerg:
Merging on `staging` so that `main` keeps having the same bytecode as the V2.1 deployments

PR #257, opened 2024-01-10
refactor: remove protocol fee checks in `MerkleStreamerLL`

### LINKED ISSUE #256: Remove fee checks in `MerkleStreamerLL`
Remove these:

- [`&& !protocolFee.gt(ud(0))`](https://github.com/sablier-labs/v2-periphery/blob/566deba675107ea88c87848fd9dea7769d4d4411/src/abstracts/SablierV2MerkleStreamer.sol#L101)
- [if (protocolFee.gt(ud(0))) ](https://github.com/sablier-labs/v2-periphery/blob/566deba675107ea88c87848fd9dea7769d4d4411/src/abstracts/SablierV2MerkleStreamer.sol#L144-L146)

As per the rationale in https://github.com/sablier-labs/company-discussions/discussions/28.

smol-ninja:
Closes #256 

## Description
This PR removes protocol fee checks in MerkleStreamerLL and allows claiming even when the protocol fee is greater than 0. When $protocolFee>0$ then ${claimAmount}=(amount - protocolFee)$.

## Changes
- **SablierV2MerkleStreamer.sol** - Removes protocol fee checks in `_checkClaim` and `clawback`.
- **Errors.sol** - Removes `SablierV2MerkleStreamer_ProtocolFeeNotZero`.

## Testing
- **claim.t.sol**: Removes `test_RevertGiven_ProtocolFeeNotZero` and adds `test_Claim_ProtocolFeeNotZero`.
- **clawback.t.sol**: Removes modifiers `givenProtocolFeeZero` and `givenProtocolFeeNotZero` as clawback does not check for protocol fee anymore 
- Updates branching trees.

#### Related: https://github.com/sablier-labs/company-discussions/discussions/28

### DIFF
--- a/src/SablierV2MerkleStreamerLL.sol
+++ b/src/SablierV2MerkleStreamerLL.sol
@@ -50,7 +50,7 @@ contract SablierV2MerkleStreamerLL is
         bool cancelable,
         bool transferable
     )
-        SablierV2MerkleStreamer(initialAdmin, asset, lockupLinear, merkleRoot, expiration, cancelable, transferable)
+        SablierV2MerkleStreamer(initialAdmin, asset, merkleRoot, expiration, cancelable, transferable)
     {
         LOCKUP_LINEAR = lockupLinear;
         streamDurations = streamDurations_;
--- a/test/integration/merkle-streamer/ll/claim/claim.t.sol
+++ b/test/integration/merkle-streamer/ll/claim/claim.t.sol
@@ -2,7 +2,7 @@
 pragma solidity >=0.8.22 <0.9.0;
 
 import { Lockup, LockupLinear } from "@sablier/v2-core/src/types/DataTypes.sol";
-import { ud } from "@prb/math/src/UD60x18.sol";
+import { ud, UD60x18 } from "@prb/math/src/UD60x18.sol";
 
 import { Errors } from "src/libraries/Errors.sol";
 
@@ -102,18 +102,16 @@ contract Claim_Integration_Test is MerkleStreamer_Integration_Test {
         _;
     }
 
-    function test_RevertGiven_ProtocolFeeNotZero()
+    function test_Claim_ProtocolFeeNotZero()
         external
         givenCampaignNotExpired
         givenNotClaimed
         givenIncludedInMerkleTree
     {
-        uint128 claimAmount = defaults.CLAIM_AMOUNT();
-        bytes32[] memory merkleProof = defaults.index1Proof();
         changePrank({ msgSender: users.admin });
         comptroller.setProtocolFee({ asset: asset, newProtocolFee: ud(0.1e18) });
-        vm.expectRevert(Errors.SablierV2MerkleStreamer_ProtocolFeeNotZero.selector);
-        merkleStreamerLL.claim({ index: 1, recipient: users.recipient1, amount: claimAmount, merkleProof: merkleProof });
+
+        test_Claim({ protocolFee: ud(0.1e18) });
     }
 
     modifier givenProtocolFeeZero() {
@@ -127,7 +125,12 @@ contract Claim_Integration_Test is MerkleStreamer_Integration_Test {
         givenIncludedInMerkleTree
         givenProtocolFeeZero
     {
+        test_Claim({ protocolFee: ud(0) });
+    }
+
+    function test_Claim(UD60x18 protocolFee) internal {
         uint256 expectedStreamId = lockupLinear.nextStreamId();
+        uint128 feeAmount = uint128(ud(defaults.CLAIM_AMOUNT()).mul(protocolFee).intoUint256());
 
         vm.expectEmit({ emitter: address(merkleStreamerLL) });
         emit Claim(defaults.INDEX1(), users.recipient1, defaults.CLAIM_AMOUNT(), expectedStreamId);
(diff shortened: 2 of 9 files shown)


andreivladbrg (changes requested):
`LOCKUP` is no longer used in abstract contract `MerkleStreamer`, thus it should be removed

https://github.com/sablier-labs/v2-periphery/blob/566deba675107ea88c87848fd9dea7769d4d4411/src/abstracts/SablierV2MerkleStreamer.sol#L37-L38

This PR should have basically undo the changes made here: https://github.com/sablier-labs/v2-periphery/pull/220

test/integration/merkle-streamer/ll/claim/claim.t.sol
```diff
@@ -127,15 +126,21 @@ contract Claim_Integration_Test is MerkleStreamer_Integration_Test {
         givenIncludedInMerkleTree
         givenProtocolFeeZero
     {
+        test_Claim(ud(0));
+    }
+
+    function test_Claim(UD60x18 protocolFee) internal {
         uint256 expectedStreamId = lockupLinear.nextStreamId();
 
+        uint128 feeAmount = uint128(ud(defaults.CLAIM_AMOUNT()).mul(protocolFee).intoUint256());
```
PaulRBerg, line 135: Let's remove the end line before this declaration

test/integration/merkle-streamer/ll/claim/claim.t.sol
```diff
@@ -127,15 +126,21 @@ contract Claim_Integration_Test is MerkleStreamer_Integration_Test {
         givenIncludedInMerkleTree
         givenProtocolFeeZero
     {
+        test_Claim(ud(0));
```
PaulRBerg, line 129: Let's use named parameters here:

```solidity
test_Claim({ protocolFee: ud(0) });
```

PaulRBerg (changes requested):
That's a beautiful PR body message, thanks for the detailed report, @smol-ninja.

Feedback below.

test/integration/merkle-streamer/ll/claim/claim.t.sol
andreivladbrg: I think this line can be removed completely, as my feedback left on `src`

src/SablierV2MerkleStreamerLL.sol
```diff
@@ -99,6 +104,8 @@ contract SablierV2MerkleStreamerLL is
         );
 
         // Log the claim.
-        emit Claim(index, recipient, amount, streamId);
+        unchecked {
+            emit Claim(index, recipient, amount - fee, streamId);
```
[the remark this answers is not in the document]
PaulRBerg: riiiiiiiiight

In this case, this is an arbitrary business decision. It's up to us if we want to use the term "claim amount" to refer to the raw amount (exclusive of fees) or to the total amount (the one inclusive of fees).

While I can make a good case in favor of either approach, I'm inclined to go with @andreivladbrg's suggestion (i.e. remove the `-fee` subtraction), because:

1. We do not expect to activate the protocol fee for the vast majority of airstreamed tokens, so the two amounts would match
2. Gas saving - this helps users

test/integration/merkle-streamer/ll/claim/claim.t.sol
PaulRBerg: yes you're right

smol-ninja: If we remove this, then the following test will not pass. Below test is necessary to test claim when `protocolFee` is non-zero and in that case amount deposited in the stream equals `claimAmount - feeAmount `.

```solidity

    function test_Claim_ProtocolFeeNotZero()
        external
        givenCampaignNotExpired
        givenNotClaimed
        givenIncludedInMerkleTree
    {
        changePrank({ msgSender: users.admin });
        UD60x18 protocolFee = ud(0.1e18);
        comptroller.setProtocolFee({ asset: asset, newProtocolFee: protocolFee });

        test_Claim(protocolFee);
    }
```

test/integration/merkle-streamer/ll/claim/claim.t.sol
smol-ninja: `protocolFee` affects the deposit amount in the lockup linear stream. When `protocolFee` is non-zero, the amount deposit will be claim amount minus fee amount. So, now we have two test functions:

```solidity
    function test_Claim_ProtocolFeeNotZero() external {
        changePrank({ msgSender: users.admin });
        UD60x18 protocolFee = ud(0.1e18);
        comptroller.setProtocolFee({ asset: asset, newProtocolFee: protocolFee });

        test_Claim(protocolFee);
    }
```

and 

```solidity

    function test_Claim() givenProtocolFeeZero {
        test_Claim(ud(0));
    }
```

And to avoid DRY, I added an internal `test_Claim` function that takes `protocolFee` as input parameter.

test/integration/merkle-streamer/ll/claim/claim.t.sol
PaulRBerg: ooh, right. you're right @smol-ninja.

smol-ninja: Thanks for confirming Paul. Do you also agree @andreivladbrg?

andreivladbrg: hmm, i am not yet sure why we need to test both versions:
- when the protocol fee = 0
- when the protocol fee > 0

isn't this already tested in the create functions in `v2-core` ?

my views to this is like if we were to test the `ERC20` balances. we are not doing this because that's a matter of `ERC20` contract tests

PaulRBerg: What you said is true in the purest technical sense but business-wise, it's a good decision to test our own system (ERC20 is different - we haven't written that ourselves).

System A might work by itself, system B might work by itself too, but when you couple A to B, things could go wrong.

so this is basically an end-to-end test and I'm in favor of keeping it just like @smol-ninja has written it.

andreivladbrg: ok got it

andreivladbrg:
@smol-ninja pushed a new commit, can you double check that the PR looks good now?

andreivladbrg (approved):

PR #259, opened 2024-01-10
ci: use reusable workflows

### LINKED ISSUE #152: Reuse GitHub workflows across repositories
Find a way to re-use GitHub workflows between this repository (V2 Periphery) and other similar repositories such as V2 Core.

For example, we could write a reusable workflow for the `lint` and the `build` jobs.

PaulRBerg:
Closes #152.

This PR reduces the complexity of our CI workflows by de-duplicating as much logic as possible. Most of the jobs in the CI workflows are now reusable, and they can be imported from this repository:

https://github.com/sablier-labs/reusable-workflows

Please look at the README of that repository for more details on what workflows are available.

### DIFF
--- a/.github/workflows/ci-deep.yml
+++ b/.github/workflows/ci-deep.yml
@@ -1,67 +1,42 @@
 name: "CI Deep"
 
-env:
-  RPC_URL_MAINNET: ${{ secrets.RPC_URL_MAINNET }}
-
 on:
   schedule:
     - cron: "0 3 * * 0" # at 3:00am UTC every Sunday
   workflow_dispatch:
     inputs:
+      integrationFuzzRuns:
+        default: "100000"
+        description: "Integration: number of fuzz runs."
+        required: false
       forkFuzzRuns:
-        default: "250"
-        description: "Number of fuzz runs for each fork test."
+        default: "1000"
+        description: "Fork: number of fuzz runs."
         required: false
 
 jobs:
   lint:
-    runs-on: "ubuntu-latest"
-    steps:
-      - name: "Check out the repo"
-        uses: "actions/checkout@v4"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-deep.yml
```diff
@@ -1,67 +1,42 @@
...
+    uses: "sablier-labs/reusable-workflows/.github/workflows/forge-lint.yml@main"
+
+  build:
+    uses: "sablier-labs/reusable-workflows/.github/workflows/forge-build.yml@main"
+
+  test-integration:
+    needs: ["lint", "build"]
+    uses: "sablier-labs/reusable-workflows/.github/workflows/forge-test.yml@main"
+    with:
+      foundry-fuzz-runs: ${{ inputs.integrationFuzzRuns || 100000 }}
```
smol-ninja, line 28: Since we have already specified `integrationFuzzRuns` above and then we also have its `default` value set in the [base-workflow](https://github.com/sablier-labs/reusable-workflows/blob/main/.github/workflows/forge-test.yml#L15C7-L15C24), isn't using `|| 100000` an overkill?

smol-ninja (changes requested):

.github/workflows/ci-deep.yml
PaulRBerg: We need both, unfortunately, because the default input values in `workflow_dispatch` are NOT taken into consideration when the cron scheduler runs the CI workflow.

https://stackoverflow.com/q/73635200/3873510

smol-ninja: I see. Thanks for explaining.

smol-ninja (approved):

PR #283, opened 2024-02-06
Update openzeppelin to v5.0.0

andreivladbrg:
The reason why I've added the the erc20 mock here to is that the package does not include the mocks `dir` from `test` in core.


smol-ninja (approved):
See comment below.

test/Base.t.sol
```diff
@@ -20,6 +19,7 @@ import { SablierV2Batch } from "src/SablierV2Batch.sol";
 import { SablierV2MerkleLockupFactory } from "src/SablierV2MerkleLockupFactory.sol";
 import { SablierV2MerkleLockupLL } from "src/SablierV2MerkleLockupLL.sol";
 
+import { ERC20Mock } from "./mocks/erc20/ERC20Mock.sol";
```
smol-ninja, line 22: > The reason why I've added the the erc20 mock here to is that the package does not include the mocks dir from test in core.

I just checked and `ERC20` mock is included in v2-core package. You can import it using 

```solidity
import { ERC20Mock } from "@sablier/v2-core/test/mocks/erc20/ERC20Mock.sol";
```

smol-ninja (changes requested):

test/Base.t.sol
andreivladbrg: The repo yes

Not the package https://www.npmjs.com/package/@sablier/v2-core?activeTab=code

We are currently installing it as a git repo but I think we are going to change this, once everything is released on core

smol-ninja: You are right. Out of curiosity, why did we decide to include only `utils` in the package? Is there a Github discussion that I can read to understand the rationale behind it?

andreivladbrg: for precompiles: https://github.com/sablier-labs/v2-core/blob/main/test/utils/Precompiles.sol

https://github.com/sablier-labs/v2-periphery/blob/8f4c3de99b56af81ef467b03d296c441f60f14f2/test/integration/Integration.t.sol#L4

smol-ninja: Got it. Thanks for explaining.

smol-ninja (approved):

smol-ninja:
Should we merge this now @andreivladbrg?

andreivladbrg:
yeah

PaulRBerg:
I have the same question as [here](https://github.com/sablier-labs/v2-core/pull/806#issuecomment-2037928508).

PR #287, opened 2024-02-07
test: add `ASSET` variable in `Fork.t.sol` to store forked asset

### LINKED ISSUE #286: forked `asset` is overridden by `Base_Test.setUp()` during fork tests
Since we are not storing forked `asset` into a separate variable, `Base_Test.setUp()` overrides the value of `asset` set by the `Fork_Test`’s constructor. That means all fork tests are operating with our own mock DAI and not with the forked asset.

[See code here ↗](https://github.com/sablier-labs/v2-periphery/blob/3e6e023623e7b0260990564047569f9e2cb2ca16/test/fork/Fork.t.sol#L19C1-L32C27)

```solidity
abstract contract Fork_Test is Base_Test, V2CoreFuzzers {
    /*//////////////////////////////////////////////////////////////////////////
                                    CONSTRUCTOR
    //////////////////////////////////////////////////////////////////////////*/
    constructor(IERC20 asset_) {
        asset = asset_;
    }

    /*//////////////////////////////////////////////////////////////////////////
                                  SET-UP FUNCTION
    //////////////////////////////////////////////////////////////////////////*/

    function setUp() public virtual override {
        // Fork Ethereum Mainnet at a specific block number.
        vm.createSelectFork({ blockNumber: 18_821_300, urlOrAlias: "mainnet" });

        // Set up the base test contract.
        Base_Test.setUp();
```

smol-ninja:
closes #286

### Summary
- Include `ffi` config in default foundry profile to prevent error while testing `Base.s.sol` locally.
- Rename `asset` to `dai` when referring to the default test asset.
- Refactoring related to tests.


PaulRBerg (approved):
looks fine

test/fork/batch/createWithMilestones.t.sol
```diff
@@ -48,12 +48,14 @@ abstract contract CreateWithMilestones_Batch_Fork_Test is Fork_Test {
         uint256 firstStreamId = lockupDynamic.nextStreamId();
         uint128 totalTransferAmount = params.perStreamAmount * params.batchSize;
 
-        deal({ token: address(asset), to: params.sender, give: uint256(totalTransferAmount) });
+        deal({ token: address(ASSET), to: params.sender, give: uint256(totalTransferAmount) });
         changePrank({ msgSender: params.sender });
-        asset.approve({ spender: address(batch), amount: totalTransferAmount });
+        // a low-level call to ignore reverts for ERC20 tokens with missing return value.
+        (bool success,) = address(ASSET).call(abi.encodeCall(IERC20.approve, (address(batch), totalTransferAmount)));
```
andreivladbrg, line 54: let's use `approveContract` function

test/fork/batch/createWithRange.t.sol
```diff
@@ -42,12 +42,14 @@ abstract contract CreateWithRange_Batch_Fork_Test is Fork_Test {
         uint256 firstStreamId = lockupLinear.nextStreamId();
         uint128 totalTransferAmount = params.perStreamAmount * params.batchSize;
 
-        deal({ token: address(asset), to: params.sender, give: uint256(totalTransferAmount) });
+        deal({ token: address(ASSET), to: params.sender, give: uint256(totalTransferAmount) });
         changePrank({ msgSender: params.sender });
-        asset.approve({ spender: address(batch), amount: totalTransferAmount });
+        // a low-level call to ignore reverts for ERC20 tokens with missing return value.
+        (bool success,) = address(ASSET).call(abi.encodeCall(IERC20.approve, (address(batch), totalTransferAmount)));
```
andreivladbrg, line 48: let's use `approveContract` function

test/Base.t.sol
```diff
@@ -247,6 +247,7 @@ abstract contract Base_Test is DeployOptimized, Events, Merkle, V2CoreAssertions
 
     function computeMerkleStreamerLLAddress(
```
andreivladbrg, line 248: should we add one more function without asset param?

```solidity 
    function computeMerkleStreamerLLAddress(
        address admin,
        bytes32 merkleRoot,
        uint40 expiration
    )
        internal
        returns (address)
    {
     computeMerkleStreamerLLAddress(admin, address(dai), merkleRoot, expiration);
    }
```

test/Base.t.sol
```diff
@@ -275,6 +276,7 @@ abstract contract Base_Test is DeployOptimized, Events, Merkle, V2CoreAssertions
 
     function getMerkleStreamerLLBytecode(
```
andreivladbrg, line 277: same above (`computeMerkleStreamerLLAddress`), should we add one more function without asset param?

andreivladbrg (changes requested):

andreivladbrg:
Note: this PR may cause a lot of conflicts on `staging`.

Should we merge this into `staging`?

smol-ninja:
> Note: this PR may cause a lot of conflicts on staging.
Should we merge this into staging?

I agree. If it's not a very critical issue (which I think it's not), even my opinion is to merge this PR against the staging branch. Do we all agree then?

test/fork/Fork.t.sol
```diff
@@ -12,12 +13,18 @@ import { Base_Test } from "../Base.t.sol";
 
 /// @notice Common logic needed by all fork tests.
 abstract contract Fork_Test is Base_Test, V2CoreFuzzers {
+    /*//////////////////////////////////////////////////////////////////////////
+                                     CONSTANTS
```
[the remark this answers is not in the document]
smol-ninja: ik. This was because the PR is against the `main` branch.

test/Base.t.sol
smol-ninja: Why? In your suggestions above, you recommended combining two functions into one using input parameters. Then why do you suggest to add one more function in this case?

andreivladbrg:
> Do we all agree then?

I personally agree

test/Base.t.sol
```diff
@@ -97,8 +97,8 @@ abstract contract Base_Test is DeployOptimized, Events, Merkle, V2CoreAssertions
     }
 
     /// @dev Labels the most relevant contracts.
-    function labelContracts() internal {
-        vm.label({ account: address(asset), newLabel: IERC20Metadata(address(asset)).symbol() });
+    function labelContracts() internal virtual {
```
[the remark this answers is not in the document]
smol-ninja: agree

smol-ninja:
I've updated this against `staging` branch. This PR may be merged independently from other PRs.

smol-ninja:
@andreivladbrg do you want to review it once more so that I can merge it into staging? It can be merged independently and I will update other PRs according to the changes introduced by this.

test/fork/assets/USDT.t.sol
```diff
@@ -7,6 +7,7 @@ import { CreateWithTimestamps_LockupDynamic_Batch_Fork_Test } from "../batch/cre
 import { CreateWithTimestamps_LockupLinear_Batch_Fork_Test } from "../batch/createWithTimestampsLL.t.sol";
 import { MerkleLockupLL_Fork_Test } from "../merkle-lockup/MerkleLockupLL.t.sol";
 
+/// @dev An ERC-20 asset that suffers from the missing return value bug.
```
andreivladbrg, line 10: Let's use other word instead of "suffers". It sound too dramatic😅

foundry.toml
```diff
@@ -3,6 +3,7 @@
   block_timestamp = 1_682_899_200 # May 1, 2023 at 00:00 GMT
   bytecode_hash = "none"
   evm_version = "paris"
+  ffi = true
```
andreivladbrg, line 6: let's exclude this temporarily, i will push a commit directly to `main`, and then we can rebase

test/Base.t.sol
andreivladbrg: to have a greater flexibility 

similar to what we have in `Defaults`. one function has the asset one has not:

https://github.com/sablier-labs/v2-periphery/blob/1804e530aabd274106655d14e4da2c7a1e39e431/test/utils/Defaults.sol#L181-L197

andreivladbrg (commented):
Looks good now

test/fork/assets/USDT.t.sol
andreivladbrg: nvm, is the same on core, we can leave it as it is

andreivladbrg:
For some reason the CI fails because `v2-core` is not installed when `bun install --frozen-lockfile` is ran

image

locally i don't face this issue:


image

test/Base.t.sol
smol-ninja: OK.

test/Base.t.sol
smol-ninja: `getMerkleLockupLLBytecode` is only used once inside `computeMerkleStreamerLLAddress`. So I think there is no need to have one more function without asset param. wdyt?

andreivladbrg: > getMerkleLockupLLBytecode is only used once

in that case, we don't need one more

PR #292, opened 2024-02-17
feat: support Blast network

smol-ninja:
Closes https://github.com/sablier-labs/v2-periphery/issues/295.

- [x] Mock Blast rebasing contracts
- [x] Update integrations tests using mocks
- [x] Update fork tests using mocks
- [x] Deploy on testnet and check configurations
- [x] Add mainnet variables
    - [x] core addresses


test/integration/merkle-streamer/ll/constructor/constructor.t.sol
```diff
@@ -80,5 +90,88 @@ contract Constructor_MerkleStreamerLL_Integration_Test is MerkleStreamer_Integra
...
+        (,,, vars.actualBlastGasMode) = blastMock.readGasParams(address(constructedStreamerLL));
+        vars.expectedBlastGasMode = GasMode.VOID;
+        assertEq(uint8(vars.actualBlastGasMode), uint8(vars.expectedBlastGasMode), "blastGasMode");
+
+        vars.actualBlastGovernor = blastMock.governorMap(address(constructedStreamerLL));
+        vars.expectedBlastGovernor = address(0);
+        assertEq(vars.actualBlastGovernor, vars.expectedBlastGovernor, "blastGovernor");
+    }
+
+    function test_ConstructorWhen_RebasingAsset() external {
```
smol-ninja, line 108: @andreivladbrg

1. Should I create another internal function for the repeat code between `test_ConstructorWhen_RebasingAsset` and `test_Constructor`?
OR
2. Should I remove the repeat code from `test_ConstructorWhen_RebasingAsset` and only keep the last 4 Blast related assertions?
OR
3. Is it fine this way?

test/fork/merkle-streamer/MerkleStreamerLL.t.sol
```diff
@@ -133,6 +135,13 @@ abstract contract MerkleStreamerLL_Fork_Test is Fork_Test {
             "MerkleStreamerLL contract does not match computed address"
         );
 
+        // Check configuration for rebasing asset on Blast L2.
```
smol-ninja, line 138: I have only kept this test here since in the case of Automatic yield, the `clawback` function can be used to claim yield which already has a test below.

All other Sablier-related functions are inherited from v2-core where we have already tested them.

Any thoughts on this?

smol-ninja:
I have removed the WETH fork contract because of the complexity in funding the caller address with WETH (foundry `deal` doesn't work on Blast because the differences between how `deal` works and how Blast `ERC20Rebasing` works). 

However, I've realized that we do not use `deal` in core contracts because we populate them with a [`HOLDER` address](https://github.com/sablier-labs/v2-core/blob/fa56458d788d90550fedd40ee25de2b3f43589c4/test/fork/assets/BWETH.t.sol#L10).

```solidity
/// @dev A WETH token with rebasing yield deployed on Blast L2.
IERC20 constant ASSET = IERC20(0x4200000000000000000000000000000000000023);
address constant HOLDER = 0x50ED0a15C0aF3CaC9A2c46FbfAAbDD09b737087C;
```

So there are two options:

1. We keep it as it is and do not use WETH in fork test of the periphery; OR
2. We introduce `Holder` in the periphery similar to v2-core design.

test/integration/merkle-streamer/ll/constructor/constructor.t.sol
andreivladbrg: I don't understand why we need these constructor tests 😅

We **are not** interacting with these contracts at creation, so we don't need to do anything.

It is like testing `ERC20` functions.

smol-ninja: Yes, you are correct. We don't need them. I wrote it when we were configuring yield in the Lockup constructor which we don't do anymore.

test/fork/merkle-streamer/MerkleStreamerLL.t.sol
smol-ninja: Same here @andreivladbrg right? Same reason as above.

src/abstracts/SablierV2MerkleStreamer.sol
```diff
@@ -5,9 +5,9 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { MerkleProof } from "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
 import { BitMaps } from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
-import { Adminable } from "@sablier/v2-core/src/abstracts/Adminable.sol";
 import { ISablierV2Lockup } from "@sablier/v2-core/src/interfaces/ISablierV2Lockup.sol";
 import { UD60x18, ud } from "@prb/math/src/UD60x18.sol";
```
andreivladbrg, line 9: let's move this import above
```solidity 
import { UD60x18, ud } from "@prb/math/src/UD60x18.sol";
import { ISablierV2Lockup } from "@sablier/v2-core/src/interfaces/ISablierV2Lockup.sol";
```

andreivladbrg (commented):
Looking good, expect for the unneeded tests

src/types/DataTypes.sol
```diff
@@ -14,12 +14,12 @@ library Batch {
...
         address sender;
-        address recipient;
-        uint128 totalAmount;
         bool cancelable;
         bool transferable;
-        LockupDynamic.SegmentWithDelta[] segments;
+        address recipient;
+        uint128 totalAmount;
         Broker broker;
+        LockupDynamic.SegmentWithDelta[] segments;
```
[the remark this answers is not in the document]
smol-ninja: Because that's the order we had in our latest release. The order was changed in [this commit](https://github.com/sablier-labs/v2-periphery/commit/52831a6b8474a0d12adb5c5929b0382c88fe2208).

src/types/DataTypes.sol
```diff
@@ -38,13 +38,13 @@ library Batch {
...
-        address recipient;
-        uint128 totalAmount;
         uint40 startTime;
         bool cancelable;
         bool transferable;
-        LockupDynamic.Segment[] segments;
+        address recipient;
+        uint128 totalAmount;
         Broker broker;
+        LockupDynamic.Segment[] segments;
```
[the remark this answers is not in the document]
smol-ninja: Same reason as above. Otherwise the ABI would not be same as the ABI we use in our latest release.

Issue #273, opened 2024-02-01
Implement "MerkleStreamerLT" (LockupTranched)


PaulRBerg:
As discussed in https://github.com/sablier-labs/company-discussions/discussions/23.

PR #297, opened 2024-03-05
MerkleLockup for LockupTranched

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/273

Simillar to Lockup Dynamic, but we store in the `MerkleLockupLT` contract an array of structs with percetanges and durations (how Razvan suggested [here](https://github.com/sablier-labs/v2-periphery/discussions/166#discussioncomment-6768109)). 

```solidty
    struct TrancheWithPercentage {
        UD2x18 amountPercentage; // 1e18 = 100%
        uint40 duration;
    }
    MerkleLockupLT.TrancheWithPercentage[] internal _tranchesWithPercentages;
```

I used `UD2x18` instead of `UD60x18` to be more gas efficient. The sum of these percentages must be equal to 100%


src/SablierV2MerkleLockupFactory.sol
```diff
@@ -51,4 +56,53 @@ contract SablierV2MerkleLockupFactory is ISablierV2MerkleLockupFactory {
...
+                baseParams.transferable,
+                lockupTranched,
+                abi.encode(tranchesWithPercentages)
+            )
+        );
+
+        // Deploy the Merkle Lockup contract with CREATE2.
+        merkleLockupLT = new SablierV2MerkleLockupLT{ salt: salt }(baseParams, lockupTranched, tranchesWithPercentages);
+
+        // Log the creation of the Merkle Lockup, including some metadata that is not stored on-chain.
```
smol-ninja, line 103: We can remove "including some metadata that is not stored on-chain" now since all the data is being stored on chain.

src/SablierV2MerkleLockupLT.sol
```diff
@@ -0,0 +1,171 @@
...
+
+        // Safe Interactions: query the protocol fee. This is safe because it's a known Sablier contract that does
+        // not call other unknown contracts.
+        UD60x18 protocolFee = LOCKUP_TRANCHED.comptroller().protocolFees(ASSET);
+
+        // Adjust the last tranche amount if the protocol fee is greater than zero.
+        // This is needed to ensure the protocol invariat: sum of tranche amounts equals the deposit amount.
+        if (protocolFee.gt(ud(0))) {
+            UD60x18 protocolFeeAmount = udAmount.mul(protocolFee);
+            tranches[trancheCount - 1].amount -= protocolFeeAmount.intoUint128();
```
smol-ninja, line 162: This works only when `tranches[trancheCount - 1].amount` > `protocolFeeAmount`. Otherwise the correct pseudo code to implement it would be

```solidity
if (tranches[trancheCount - 1].amount > protocolFeeAmount.intoUint128()) {
    tranches[trancheCount - 1].amount -= protocolFeeAmount.intoUint128();
} else if (tranches[trancheCount - 2].amount > protocolFeeAmount.intoUint128() - tranches[trancheCount - 1].amount) {
    tranches[trancheCount - 2].amount -= (protocolFeeAmount.intoUint128() - tranches[trancheCount - 1].amount);
} else if 
.......
```

smol-ninja (changes requested):
Reviewed src. [Made a PR](https://github.com/sablier-labs/v2-periphery/pull/301) to arrange imports alphabetically.

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Done in https://github.com/sablier-labs/v2-periphery/pull/301/commits/c1c7b0bb699c2bde672c86fb4b4cdb498db84e09

andreivladbrg: not all data is stored on chain:

`aggregateAmount` and `recipientsCount` isn't

src/types/DataTypes.sol
```diff
@@ -83,3 +85,17 @@ library MerkleLockup {
...
+library MerkleLockupLT {
+    /// @notice Struct encapsulating the amount percentage and the tranche duration of the stream.
+    /// @dev Each recipient may have a different amount allocated, this struct stores the percentage of the
+    /// amount designated for each duration unlock. We use a 18 decimals format to represent percentages:
+    /// 100% = 1e18.
+    /// @param amountPercentage The percentage of the amount designated to be unlocked in this tranche.
+    /// @param duration The time difference in seconds between this tranche and the previous one.
+    struct TrancheWithPercentage {
+        // slot 0
+        UD2x18 amountPercentage;
```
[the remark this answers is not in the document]
andreivladbrg: hmm, let me think about it

src/SablierV2MerkleLockupLT.sol
andreivladbrg: correct, so the better version would be to:
- calculate deposit amount: `totalAmount - protocolFeeAmount`
- calculate tranches amounts from `depositAmount` instead of `totalAmount`

test/utils/Assertions.sol
```diff
@@ -0,0 +1,26 @@
...
+    /// @dev Compares two {MerkleLockupLT.TrancheWithPercentage[]} arrays.
+    function assertEq(
+        MerkleLockupLT.TrancheWithPercentage[] memory a,
+        MerkleLockupLT.TrancheWithPercentage[] memory b
+    )
+        internal
+    {
+        if (keccak256(abi.encode(a)) != keccak256(abi.encode(b))) {
+            emit Log("Error: a == b not satisfied [MerkleLockupLT.TrancheWithPercentage[]]");
+            emit LogNamedArray("   Left", b);
```
smol-ninja, line 21: Why do we emit "Left" key with right array?

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Yes you are correct. I was focussing on ipfsCID only.

test/integration/merkle-lockup/factory/create-merkle-lockup-lt/createMerkleLockupLT.tree
```diff
@@ -0,0 +1,15 @@
+createMerkleLockupLL.t.sol
+├── when the tranches percentages sum does not equal 100%
+│   ├── when the tranches percentages sum is less than 100%
```
smol-ninja, line 3: can we not just say

- when the tranches percentages sum does not equal 100%
    - it should revert
- when the tranches percentages sum equals 100%
.......

test/integration/merkle-lockup/factory/create-merkle-lockup-lt/createMerkleLockupLT.t.sol
```diff
@@ -0,0 +1,159 @@
...
+import { MerkleLockup, MerkleLockupLT } from "src/types/DataTypes.sol";
+
+import { MerkleLockup_Integration_Test } from "../../MerkleLockup.t.sol";
+
+contract CreateMerkleLockupLT_Integration_Test is MerkleLockup_Integration_Test {
+    function setUp() public override {
+        MerkleLockup_Integration_Test.setUp();
+    }
+
+    modifier whenTranchesPercentageSumIsNotOneHundred() {
```
smol-ninja, line 17: If we go with `unlockPercentage` then we can call it `whenUnlockPercentageSumisNot100`? Thats still too long 😅

test/integration/merkle-lockup/lt/get-tranches-with-percentage/getTranchesWithPercentage.t.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22 <0.9.0;
+
+import { MerkleLockupLT } from "src/types/DataTypes.sol";
+
+import { MerkleLockup_Integration_Test } from "../../MerkleLockup.t.sol";
+
+contract GetTranchesWithPercentages_Integration_Test is MerkleLockup_Integration_Test {
```
smol-ninja, line 8: We dont need this test. We are already testing `getTranchesWithPercentages` in `Constructor_MerkleLockupLT_Integration_Test`

smol-ninja (changes requested):
I have reviewed the PR. Left my comments below.

I think we should find a new name for tranches percentages. `tranchesWithPercentages` does not look apt to me.

test/fork/merkle-lockup/MerkleLockupLT.t.sol
```diff
@@ -0,0 +1,194 @@
...
+        Vars memory vars;
+        vars.recipientsCount = params.leafData.length;
+        vars.amounts = new uint128[](vars.recipientsCount);
+        vars.indexes = new uint256[](vars.recipientsCount);
+        vars.recipients = new address[](vars.recipientsCount);
+        for (uint256 i = 0; i < vars.recipientsCount; ++i) {
+            vars.indexes[i] = params.leafData[i].index;
+
+            // Bound each leaf amount so that `aggregateAmount` does not overflow.
+            params.leafData[i].amount =
```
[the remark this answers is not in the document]
andreivladbrg: correct

andreivladbrg:
> I think we should find a new name for tranches percentages. tranchesWithPercentages does not look apt to me

I agree that the name is not the greatest, but I couldn't find something better. Do you have other suggestions?

test/utils/Assertions.sol
andreivladbrg: nice finding, this is a mistake which has been done in the core as well: 

https://github.com/sablier-labs/v2-core/blob/main/test/utils/Assertions.sol

src/SablierV2MerkleLockupLT.sol
andreivladbrg: in the end, we are going to remove the fees, so this handle is not going to be necessary

test/utils/Assertions.sol
smol-ninja: Correct. Fixed in https://github.com/sablier-labs/v2-core/pull/838.

test/integration/merkle-lockup/MerkleLockup.t.sol
```diff
@@ -41,6 +53,22 @@ abstract contract MerkleLockup_Integration_Test is Integration_Test {
         return computeMerkleLockupLLAddress(admin, merkleRoot, defaults.EXPIRATION());
     }
 
+    function computeMerkleLockupLTAddress() internal returns (address) {
```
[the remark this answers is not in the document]
andreivladbrg: we could

test/integration/merkle-lockup/factory/create-merkle-lockup-lt/createMerkleLockupLT.tree
andreivladbrg: we could, be it's better to handle to cover all cases: when less and greater

src/types/DataTypes.sol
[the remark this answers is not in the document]
andreivladbrg: made the change in the [commit](https://github.com/sablier-labs/v2-periphery/pull/297/commits/0d123007ca6a78357fa141da1cb35668bf5410b0)

andreivladbrg:
@smol-ninja addressed all the feedback, lmk if it looks good.

for some reason the ci passes, still not sure why

smol-ninja:
Every time there is a new commit in v2-core#staging, the bun lockfile needs to be updated to fetch the latest commit. Without which, ci fails. Not sure why this behaviour though.

smol-ninja (changes requested):
Looks good to me except the following comments:

> Also, should we not use plural?

I thought about it. My view is that using a plural for the variable name makes more sense. Struct `TrancheWithPercentage` represents single tranche while `tranchesWithPercentages` represents a set of tranche. So I am good with using plural in this context. However I have the following suggestions:

1. Because it leads to longer names, how about we use synonym of percentage which is Pct - [see Cambridge definition](https://dictionary.cambridge.org/dictionary/english/pct)? That will change `tranchesWithPercentages` to `tranchesWithPct`. I know it might not be super clear compared to the original name, but curious to know what your think about it.

For example,
- `actualTranchesWithPercentages` becomes `actualTranchesWithPct`.
- `MerkleLockupLT.TrancheWithPercentage[] tranchesWithPercentages` becomes `MerkleLockupLT.TrancheWithPct[] tranchesWithPct`.
- `SablierV2MerkleLockupFactory_PercentageSumNotEqualOneHundred` becomes `SablierV2MerkleLockupFactory_TotalPctNotEqual100`

2. And instead of `amountsSum` and `PercentagesSum`, how about `totalAmount` and `totalPercentage` / `totalPct`? But If double `s` doesn't look weird to you, feel free to ignore this and such comments.

src/SablierV2MerkleLockupLT.sol
```diff
@@ -0,0 +1,160 @@
...
+    function _calculateTranches(uint128 amount)
+        internal
+        view
+        returns (LockupTranched.TrancheWithDuration[] memory tranches)
+    {
+        // Load the tranches in memory to save gas.
+        MerkleLockupLT.TrancheWithPercentage[] memory tranchesWithPercentages = _tranchesWithPercentages;
+
+        // Declare the variables need for calculation.
+        UD60x18 trancheAmountsSum;
```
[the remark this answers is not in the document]
andreivladbrg: for this i would like to be consistent, in `core` we use double `s`

https://github.com/sablier-labs/v2-core/blob/4b3b51562b27f2d7cbf294f332db8f8c3689f1d9/src/libraries/Helpers.sol#L240

https://github.com/sablier-labs/v2-core/blob/4b3b51562b27f2d7cbf294f332db8f8c3689f1d9/src/libraries/Helpers.sol#L302

test/integration/merkle-lockup/factory/create-merkle-lockup-lt/createMerkleLockupLT.t.sol
```diff
@@ -0,0 +1,159 @@
...
+contract CreateMerkleLockupLT_Integration_Test is MerkleLockup_Integration_Test {
+    function setUp() public override {
+        MerkleLockup_Integration_Test.setUp();
+    }
+
+    modifier whenPercentagesSumIsNotOneHundred() {
+        _;
+    }
+
+    function test_RevertWhen_PercentageSumLessThanOneHundred() external whenPercentagesSumIsNotOneHundred {
```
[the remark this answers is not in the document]
andreivladbrg: i did this because the branch test tree has these:

https://github.com/sablier-labs/v2-periphery/blob/048894d0a339e2028fb868afd42e1050639e4a82/test/integration/merkle-lockup/factory/create-merkle-lockup-lt/createMerkleLockupLT.tree#L2-L7

src/SablierV2MerkleLockupLT.sol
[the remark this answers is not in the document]
smol-ninja: Fair enough.

smol-ninja (approved):
LGTM except minor refactors in `createMerkleLockupLT.t.sol` and `createMerkleLockupLT.tree`.

andreivladbrg:
> My view is that using a plural for the variable name makes more sense. Struct TrancheWithPercentage represents single tranche while tranchesWithPercentages represents a set of tranche.

My reasoning was the same.

> Because it leads to longer names, how about we use synonym of percentage which is Pct - [see Cambridge definition](https://dictionary.cambridge.org/dictionary/english/pct)? That will change tranchesWithPercentages to tranchesWithPct. I know it might not be super clear compared to the original name, but curious to know what your think about it.

Interesting idea, but the name is not that long. Plus, the core structs have pretty much the same size (e.g., `SegmentWithDuration`).

Also, I discussed it with Paul, and he said that it is clearer with the long form (Pct might be confusing to some individuals who are not familiar with the term).

>Please review this commit: https://github.com/sablier-labs/v2-periphery/commit/6ad73edfc4430cb7ba4e347f09c9cb406dae10bb

It looks good, thanks

andreivladbrg:
> LGTM except minor refactors in createMerkleLockupLT.t.sol and createMerkleLockupLT.tree.

forced push a commit, lmk if everything is alright, if yes, can you approve the PR?

smol-ninja (approved):

test/integration/merkle-lockup/lt/claim/claim.t.sol
```diff
@@ -101,7 +109,69 @@ contract Claim_Integration_Test is MerkleLockup_Integration_Test {
...
+        uint256 aggregateAmount = defaults.CLAIM_AMOUNT() * 3 + uint256(claimAmount);
+
+        // Compute the Merkle tree.
+        leaves = defaults.getLeaves();
+        leaves[0] = MerkleBuilder.computeLeaf(defaults.INDEX1(), users.recipient1, claimAmount);
+        MerkleBuilder.sortLeaves(leaves);
+        bytes32 merkleRoot = getRoot(leaves.toBytes32());
+
+        // Compute the Merkle proof.
+        uint256 leaf = MerkleBuilder.computeLeaf(defaults.INDEX1(), users.recipient1, claimAmount);
```
smol-ninja, line 132: You can also define this `leaf` above the line in which you substitute `leaves[0]` (L127). Then you can substitute using `leaves[0] = leaf`. This avoids computing leaf two times.

test/integration/merkle-lockup/lt/claim/claim.t.sol
```diff
@@ -101,7 +109,69 @@ contract Claim_Integration_Test is MerkleLockup_Integration_Test {
...
+
+    function test_Claim_TrancheAmountCalculationRoundingError()
+        external
+        givenCampaignNotExpired
+        givenNotClaimed
+        givenIncludedInMerkleTree
+    {
+        // Declare an amount that will cause a rounding error.
+        uint128 claimAmount = 340_282_366_920_938_463_463_374_607_431_768_211_453;
+        uint256 aggregateAmount = defaults.CLAIM_AMOUNT() * 3 + uint256(claimAmount);
```
smol-ninja, line 123: If we use 10,001, then it becomes 

```solidity
uint256 aggregateAmount = defaults.AGGREGATE_AMOUNT() + 1;
```

smol-ninja (changes requested):

src/SablierV2MerkleLockupLT.sol
```diff
@@ -0,0 +1,160 @@
...
+
+            // Assign calculated amount and duration to the tranche.
+            tranches[i] = LockupTranched.TrancheWithDuration({
+                amount: trancheAmount.intoUint128(),
+                duration: tranchesWithPercentages[i].duration
+            });
+        }
+
+        // Adjust the last tranche amount to prevent claim failure due to rounding differences during calculations. We
+        // need to ensure the protocol invariant: the sum of all tranches' amounts equals the deposit amount.
```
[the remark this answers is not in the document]
andreivladbrg: since we are going to remove the protocol fee, it is now correct to use "claim" instead of "deposit", before it wasn't because `claim > deposit`

test/integration/merkle-lockup/lt/claim/claim.t.sol
```diff
@@ -101,7 +109,69 @@ contract Claim_Integration_Test is MerkleLockup_Integration_Test {
...
+    uint256[] public leaves = new uint256[](4); // same number of recipients as in the defaults
+
+    function test_Claim_TrancheAmountCalculationRoundingError()
+        external
+        givenCampaignNotExpired
+        givenNotClaimed
+        givenIncludedInMerkleTree
+    {
+        // Declare an amount that will cause a rounding error.
+        uint128 claimAmount = 340_282_366_920_938_463_463_374_607_431_768_211_453;
```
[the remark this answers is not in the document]
andreivladbrg: this number was found by the fork fuzz test, so i used it 😅

i agree, we can change to simpler number

test/integration/merkle-lockup/lt/claim/claim.t.sol
```diff
@@ -101,7 +109,69 @@ contract Claim_Integration_Test is MerkleLockup_Integration_Test {
         _;
     }
 
-    function test_Claim() external givenCampaignNotExpired givenNotClaimed givenIncludedInMerkleTree {
+    // Needed this variables in storage due to how the imported libaries work.
+    uint256[] public leaves = new uint256[](4); // same number of recipients as in the defaults
```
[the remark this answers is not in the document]
andreivladbrg: I thought about it, but we only need this in `test_Claim_TrancheAmountsSumNotEqualClaimAmount`

test/integration/merkle-lockup/lt/claim/claim.t.sol
andreivladbrg: we can even remove this line and use `leaves[0]`

andreivladbrg:
thanks for the review @smol-ninja 

i've pushed a commit now, lmk if it looks good

src/SablierV2MerkleLockupLT.sol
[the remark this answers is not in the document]
smol-ninja: I think even back then, calling it deposit amount doesn't look apt. In the presence of protocol fee, this would have been `claimAmount - protocolFee` which is still not the deposit amount.

Also, in your last commit, I don't see this change.

smol-ninja (approved):
LGTM except my last reply.

test/integration/merkle-lockup/lt/claim/claim.t.sol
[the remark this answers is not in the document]
smol-ninja: Oh! 😅

Issue #299, opened 2024-03-06
Add batch create functions for LockupTranched


andreivladbrg:
We need to add the batch functions for `LockupTranched` in `SablierV2Batch`.

PR #300, opened 2024-03-07
Feat/add tranched batches

andreivladbrg:
Closes https://github.com/sablier-labs/v2-periphery/issues/299

### Note: 
This PR also upgrades the core module to a version that no longer has the protocol fee - thus some updates here were needed as well.


andreivladbrg:
Thanks for the review. I just removed the unused functions.

PR #302, opened 2024-03-09
Add precompiles to npm package

### LINKED ISSUE #238: Use precompiles to write a protocol deployment script
Write a deployment script that uses the V2 Core precompiles to deploy the entire Sablier V2 protocol (V2 Core + V2 Periphery).

This is possible to do now because the precompiles are part of the [`@sablier/v2-core`](https://www.npmjs.com/package/@sablier/v2-core?activeTab=code) npm package (in the `test/utils` package).

Note: there is an appearance of misuse by importing production code from `test/utils.` As explained [here](https://github.com/sablier-labs/v2-core/discussions/662), we should consider moving the precompiles - perhaps in `script`, or in `artifacts`, or in `src.

Related: https://github.com/sablier-labs/v2-periphery/pull/91#issuecomment-1580391265

smol-ninja:
Closes #238. Depends on https://github.com/sablier-labs/v2-core/pull/841 being merged first.


smol-ninja:
@andreivladbrg lmk if it looks good. I've no idea why build is failing in the ci, its working fine locally.

script/DeployPrecompiles.s.sol
andreivladbrg, line 1: why do we need this file?

why can't we import and use `Base_Script`?

smol-ninja: If my understanding is correct, the goal is to have a deployment script that can deploy the entire protocol using precompiles. This script should be included in the npm package.

If we import `Base_Script`, then we will also have to include `Base_Script` in the npm package.

References:
- https://github.com/sablier-labs/v2-periphery/issues/238
- https://github.com/sablier-labs/v2-core/issues/789

smol-ninja: I like the idea. I don't see any issue with that.

test/utils/Defaults.sol
```diff
@@ -29,7 +29,7 @@ contract Defaults is Merkle {
     uint40 public immutable CLIFF_TIME;
     uint40 public immutable END_TIME;
     uint256 public constant ETHER_AMOUNT = 10_000 ether;
-    uint256 public constant MAX_SEGMENT_COUNT = 1000;
+    uint256 public constant MAX_COUNT = 500;
```
smol-ninja, line 32: I have renamed `MAX_SEGMENT_COUNT` to `MAX_COUNT`. `MAX_SEGMENT_COUNT` is not used anywhere.

test/utils/Precompiles.t.sol
```diff
@@ -1,15 +1,20 @@
...
 
+import { Precompiles } from "../../precompiles/Precompiles.sol";
 import { ISablierV2Batch } from "../../src/interfaces/ISablierV2Batch.sol";
 import { ISablierV2MerkleLockupFactory } from "../../src/interfaces/ISablierV2MerkleLockupFactory.sol";
 
 import { Base_Test } from "../Base.t.sol";
-import { Precompiles } from "./Precompiles.sol";
 
-contract Precompiles_Test is Base_Test {
+contract Precompiles_Test is Base_Test, V2DeployOptimized {
```
smol-ninja, line 17: @andreivladbrg do you have any suggestion on how can I re-use `DeployOptimized` from core? Should I import it in `DeployOptimized` of periphery?

script/DeployPrecompiles.s.sol
smol-ninja: As discussed privately, we decided to not write `DeployPrecompiles` script, because it exceeds EVM gas limit when using it for the deployment of the entire protocol.

test/utils/Precompiles.t.sol
andreivladbrg: I think importing it here would be alright - tho you can rename it to `V2CoreDeployOptimized`

smol-ninja: So I have removed the test `test_DeployProtocol` until we have a method to test it properly.

andreivladbrg: hmm, that's weird, the `out-optimized` dir should contain the core contracts compiled as well since we import them

at any rate, the `deployProtocol() ` function calls the `deployCore()` and `deployPeriphery` functions which are individually tested, so it should be fine

this problem, and many more, are going to be fixed when address the "Package tethering" issue

smol-ninja: Yes you are correct. If no more changes required, can you please approve this?

andreivladbrg: just brainstorming

should we use `deployProtocol` in `Base_Test` instead having to diferent functions:
- `deployPeripheryConditionally()`
- `deployDependencies()`

smol-ninja: hmm, good idea. 

Wouldn't we have to update precompiles during development every time we make changes to src contracts? `deployProtocol` uses precompiles while the current setup uses the build contracts.

andreivladbrg: > Wouldn't we have to update precompiles during development every time we make changes to src contracts?

correct, so we can leave it as it is

andreivladbrg (approved):

PR #305, opened 2024-03-18
Integrate sphinx for deployment

smol-ninja:
Same as https://github.com/sablier-labs/v2-core/pull/840


script/Base.s.sol
```diff
@@ -45,6 +54,19 @@ contract BaseScript is Script {
         vm.stopBroadcast();
     }
 
+    /// @dev Configures the Sphinx plugin to use Sphinx managed deployment for smart contracts.
+    /// Refer to https://github.com/sphinx-labs/sphinx/tree/main/docs.
+    /// CLI example:
+    /// - bun sphinx propose script/DeployBatch.s.sol --networks testnets --sig "runSphinx()"
+    function configureSphinx() public override {
+        sphinxConfig.mainnets = ["arbitrum", "avalanche", "bnb", "gnosis", "ethereum", "optimism", "polygon"];
```
andreivladbrg, line 62: Is `Base` chain not supported?

andreivladbrg (approved):
lgtm

script/Base.s.sol
smol-ninja: Yes it is. Each chain, mentioned in `foundry.toml`, should be added here. Since there was no `Base`, I didn't add `Base` here.

andreivladbrg: ok, got it. since there is traction to base L2, i would add it in the `foundry.toml` file

andreivladbrg:
the CI is failing due to new commits to `staging` branch in core, lets first merge https://github.com/sablier-labs/v2-core/pull/864 and then update this

smol-ninja:
Yes looks good to me.

PR #309, opened 2024-03-28
feat: include stream duration in CreateMerkleLockupLT event

smol-ninja:
As discussed in https://github.com/sablier-labs/v2-periphery/discussions/308

1. Include total duration of stream in `CreateMerkleLockupLT` event. (+805 gas)
2. Remove `indexed` from `baseParams` in `CreateMerkleLockupLT` and `CreateMerkleLockupLL` events. (+2466 gas primarily because of `name` and `ipfsCID` strings)

As discussed privately on slack, the rationale behind removing `indexed` from `baseParams` is that the [topics](https://docs.soliditylang.org/en/v0.8.25/contracts.html#events) can't hold more than 32 bytes. Thus, indexed tuples are hashed before emitted as a topics making them impossible to decode.


src/SablierV2MerkleLockupFactory.sol
```diff
@@ -68,12 +68,16 @@ contract SablierV2MerkleLockupFactory is ISablierV2MerkleLockupFactory {
         external
         returns (ISablierV2MerkleLockupLT merkleLockupLT)
     {
-        // Calculate the sum of percentages across all tranches.
+        // Calculate the sum of percentages and durations across all tranches.
         UD60x18 totalPercentage;
-        uint256 trancheCount = tranchesWithPercentages.length;
-        for (uint256 i = 0; i < trancheCount; ++i) {
+        uint256 totalDuration;
+        for (uint256 i = 0; i < tranchesWithPercentages.length; ++i) {
```
smol-ninja, line 74: Since there is no room for new variables on stack, I've removed `trancheCount` variable. I've also found, by running tests, that is more gas efficient (by 6-7 units).

andreivladbrg: This is partially true. For a `count` that is greater, this is not true, due to `tranchesWithPercentages.length` being accessed multiple times.

For example you can test this contract on remix:

<details>
<summary> Test contract </summary>

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.25;

contract Test {
    uint[] public ids;  

    // 12044
    function testLoopOutside() public view {
        uint count = ids.length;
        uint sum;
        for (uint i = 0; i < count; ++i){
            sum += i;
        }
    }

    // 16333
    function testLoopInside() public view {
        uint sum;
        for (uint i = 0; i < ids.length; ++i){
            sum += i;
        }
    }

    function set() public {
        for (uint i = 0; i < 40; i++){
            ids.push(10);
        }
    }
}

```
</details>

src/SablierV2MerkleLockupFactory.sol
```diff
@@ -68,12 +68,16 @@ contract SablierV2MerkleLockupFactory is ISablierV2MerkleLockupFactory {
...
             UD60x18 percentage = (tranchesWithPercentages[i].unlockPercentage).intoUD60x18();
             totalPercentage = totalPercentage.add(percentage);
+            unchecked {
+                // safe to use `unchecked` because its only used in the event.
```
andreivladbrg, line 78: we could use capital letter

andreivladbrg (changes requested):
generally looks good

left two comments

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Yes you are right. When a count is large enough, such as 100, having `trancheCount` is cheaper by 287. Just a small note, in your case, you are accessing length of a storage array, thats why the difference is so large, whereas, in our case, its memory.

Is it OK to go ahead with not having `trancheCount` since it only makes it expensive by 287? If we add it, then we wont be able to add `totalDuration`. Wdyt?

<details>
<summary> Test contract </summary>

I used the following code.

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity >=0.8.22 <0.9.0;

import { UD2x18 } from "@prb/math/src/UD2x18.sol";
import { console2 } from "forge-std/src/console2.sol";
import { MerkleLockup, MerkleLockupLT } from "src/types/DataTypes.sol";

import { Integration_Test } from "./integration/Integration.t.sol";

contract CreateMerkleLT_Gas_Test is Integration_Test {
    uint128 private constant PER_TRANCHE_AMOUNT = 1000e18;

    function setUp() public override {
        super.setUp();

        deal({ token: address(dai), to: users.alice, give: type(uint256).max });
        changePrank({ msgSender: users.alice });
        dai.approve(address(merkleLockupFactory), type(uint256).max);
    }

    function test_GasCreateMerkleLockupLT() public {
        MerkleLockup.ConstructorParams memory baseParams = MerkleLockup.ConstructorParams({
            initialAdmin: users.alice,
            asset: dai,
            ipfsCID: "1",
            name: "1",
            merkleRoot: bytes32(0x000000),
            expiration: uint40(block.timestamp + 4 weeks),
            cancelable: true,
            transferable: true
        });

        MerkleLockupLT.TrancheWithPercentage[] memory trancheWithPct = new MerkleLockupLT.TrancheWithPercentage[](100);

        for (uint256 i; i < 100; ++i) {
            trancheWithPct[i] = MerkleLockupLT.TrancheWithPercentage({
                unlockPercentage: UD2x18.wrap(0.01e18),
                duration: uint40(1 weeks)
            });
        }

        uint256 beforeGas = gasleft();
        merkleLockupFactory.createMerkleLockupLT({
            baseParams: baseParams,
            lockupTranched: lockupTranched,
            tranchesWithPercentages: trancheWithPct,
            aggregateAmount: PER_TRANCHE_AMOUNT * 100,
            recipientsCount: 5
        });
        uint256 afterGas = gasleft();

        console2.log("%s gas", beforeGas - afterGas);
    }
}

// gas usage
// - without trancheCount: 4792935
// - add trancheCount: 4792648
```

</details>

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Agree.

src/SablierV2MerkleLockupFactory.sol
andreivladbrg: > Is it OK to go ahead with not having trancheCount since it only makes it expensive by 287? If we add it, then we wont be able to add totalDuration. Wdyt?

yeah, it is a small difference, and imo it is ok to leave it as is`for (uint256 i = 0; i < tranchesWithPercentages.length; ++i) `

PR #315, opened 2024-04-09
refactor: import BaseScript from core

smol-ninja:


PaulRBerg:
Thanks for this PR, but I will close it because I am against including the `BaseScript` in the npm package.

PR #325, opened 2024-04-11
ci: remove deployment workflow

PaulRBerg:
Similar to https://github.com/sablier-labs/v2-core/pull/816.


smol-ninja (approved):

smol-ninja:
We can merge it without checks passing. Its just bun lock upgrade.

PR #335, opened 2024-04-30
refactor: `Range` to `Timestamp`

smol-ninja:
Related: https://github.com/sablier-labs/v2-core/pull/913


PaulRBerg (changes requested):
The singular `timestamp` should be `timestamps`.

andreivladbrg (approved):

Issue #353, opened 2024-06-17
Switch back to CREATE2 for creating the MerkleLockup campaigns


PaulRBerg:
## Problem

The use of `CREATE` can be exploited when the blockchain is reorged. For more details, see [M-04. Use of CREATE method is suspicious of reorg attack](https://www.codehawks.com/report/clvb9njmy00012dqjyaavpl44#M-04) in CodeHawks' audit report.

### My Thoughts

The odds for this vulnerability to occur are very low because stealing of funds can occur only in an exquisite set of circumstances:

- Alice creates airstream campaign
- Alice sends another tx to fund the campaign
    - Note: contrary to what Rusty said [here](https://github.com/cantinasec/review-sablier-3/issues/1#issuecomment-2156787774), this issue applies even when Alice uses the same wallet for the funding tx because the higher-nonce tx may still be kept in the mempool
- Alice's funding tx remains in the mempool WITHOUT being included in the previously reorged blockchain
- Eve notices this setup and sends a tx to create an airstream campaign, sniping Alice's campaign contract
- Alice's funding tx gets processed now and Eve clawbacks

## Solution

- [x] Switch to CREATE2
- [x] Remove percentage sum check in create function and move it to the claim function
- [x] Add appropriate tests

## Related

Rusty's finding during the private audit: https://github.com/cantinasec/review-sablier-3/issues/1

smol-ninja:
Instead of generating salt using just input parameters, how about we also add the followings to create2 salt:

1. `block.timestamp`: all input parameters being same, `block.timestamp` is the only value that guarantees to be cahnged over time.

2. `block.chainid`: Prevents cross-chain replay attacks. Re [L-05](https://www.codehawks.com/report/clvb9njmy00012dqjyaavpl44#L-05) in case the same Airstream is deployed on multiple networks.

Doesn't add much gas, but lowers the chances of address collision.

smol-ninja:
Also, as a solution to this [issue related to `createMerkleLT`](https://github.com/cantinasec/review-sablier-3/issues/1), how about the following:

1. Exclude `tranchesWithPercentages` from `salt`. So there is no need to revert in `createMerkleLT` execution.
2. Store the total percentage sum as a `CONSTANT` in `SablierV2MerkleLT` contract.
3. `claim` reverts if the constant value is not 100%. 

Because its a constant, it would not add much gas overhead and at the same time, it prevents funds being stuck (counterfactual deployment) in case the percentages dont add up.

[...]

Issue #327, opened 2024-04-12
Replace `computeCreate2Address` (deprecated) with `vm.computeCreate2Address`


smol-ninja:
<img width="855" alt="Screenshot 2024-04-12 at 14 49 53" src="https://github.com/sablier-labs/v2-periphery/assets/6676622/5200e0f9-af72-4c08-aaaa-c920f48ed7fd">

ThreeHrSleep:
can take this on :)

PR #339, opened 2024-05-03
refactor: use `CREATE` instead of `CREATE2`

smol-ninja:
Fixes #337.

PR #355, opened 2024-06-17
feat: use create2 to deploy merkle lockup campaigns

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/v2-periphery/issues/353
- Closes https://github.com/sablier-labs/v2-periphery/issues/327

### Note:
- Most of the changes are reversal of https://github.com/sablier-labs/v2-periphery/pull/339.


src/SablierV2MerkleLockupFactory.sol
```diff
@@ -1,22 +1,108 @@
...
 
 /// @title SablierV2MerkleLockupFactory
 /// @notice See the documentation in {ISablierV2MerkleLockupFactory}.
 contract SablierV2MerkleLockupFactory is ISablierV2MerkleLockupFactory {
+    /*//////////////////////////////////////////////////////////////////////////
+                            USER-FACING CONSTANT FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice inheritdoc ISablierV2MerkleLockupFactory
+    function computeMerkleLL(
```
PaulRBerg, line 23: I changed my mind about these utilities. They are verbose, and they might encourage users to prefund the campaigns, whereas our purpose is to get rid of counterfactual usage.

I will remove them.

src/SablierV2MerkleLockupFactory.sol
```diff
@@ -32,8 +118,24 @@ contract SablierV2MerkleLockupFactory is ISablierV2MerkleLockupFactory {
         external
         returns (ISablierV2MerkleLL merkleLL)
     {
-        // Deploy the MerkleLockup contract with CREATE.
-        merkleLL = new SablierV2MerkleLL(baseParams, lockupLinear, streamDurations);
+        // Hash the parameters to generate a salt.
+        bytes32 salt = keccak256(
+            abi.encodePacked(
+                baseParams.asset,
```
PaulRBerg, line 124: We also need to include the `msg.sender` to prevent cross-chain replays.

Unlikely to be an issue because the `initialAdmin` is part of the salt, but it shouldn't hurt to add this as an additional precautionary measure.

src/SablierV2MerkleLockupFactory.sol
```diff
@@ -62,13 +164,25 @@ contract SablierV2MerkleLockupFactory is ISablierV2MerkleLockupFactory {
...
+                baseParams.cancelable,
+                baseParams.expiration,
+                baseParams.initialAdmin,
+                abi.encode(baseParams.ipfsCID),
+                baseParams.merkleRoot,
+                bytes32(abi.encodePacked(baseParams.name)),
+                baseParams.transferable,
+                lockupTranched,
+                totalPercentage
```
PaulRBerg, line 179: There's no value in including the total percentage in the salt. In the vast majority of cases, it will be 100%, so it won't add any differentiation.

PaulRBerg: A corollary of not needing the percentage for the salt is that we can move the calculations in the MerkleLT contract's constructor.

PaulRBerg (approved):
feedback below, already addressed in my commits.

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Good point

src/SablierV2MerkleLockupFactory.sol
smol-ninja: It would create two `for` loops in `createMerkleLT` execution. Also if we do not include `totalPercentage` in `salt`, should we include `tranchesWithPercentages`? All parameters being equal, two airstreams with different `tranchesWithPercentages` would lead to collision.

smol-ninja: nvm. There were already two `for` loops.

smol-ninja: but thoughts on adding `tranchesWithPercentages` in `salt`?

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Addressed in https://github.com/sablier-labs/v2-periphery/pull/355/commits/fa7eb777b4b0597af7827c21a92394de8a45e327

src/SablierV2MerkleLockupFactory.sol
andreivladbrg: since we include `streamDurations` in LL salt, we can also include `tranchesWithPercentages` in LT salt

src/SablierV2MerkleLT.sol
```diff
@@ -46,12 +51,14 @@ contract SablierV2MerkleLT is
...
 
-        // Since Solidity lacks a syntax for copying arrays of structs directly from memory to storage, a manual
-        // approach is necessary. See https://github.com/ethereum/solidity/issues/12783.
-        uint256 count = tranchesWithPercentages.length;
-        for (uint256 i = 0; i < count; ++i) {
+        // Calculate the total percentage of the tranches and save them in the contract storage.
+        uint64 totalPercentage;
+        for (uint256 i = 0; i < tranchesWithPercentages.length; ++i) {
+            uint64 percentage = tranchesWithPercentages[i].unlockPercentage.unwrap();
+            totalPercentage = totalPercentage + percentage;
```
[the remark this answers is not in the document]
smol-ninja: Yeah we should do that.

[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/v2-periphery/pull/355/commits/321a6da017d6ba39964859cca44dcc3f862e57fa

src/SablierV2MerkleLockupFactory.sol
smol-ninja: Since there is no objection: https://github.com/sablier-labs/v2-periphery/pull/355/commits/321a6da017d6ba39964859cca44dcc3f862e57fa

PaulRBerg: happy to include `tranchesWithPercentages` in the salt

PaulRBerg:
OK, looks good to me now

src/SablierV2MerkleLT.sol
```diff
@@ -46,12 +51,14 @@ contract SablierV2MerkleLT is
     {
         LOCKUP_TRANCHED = lockupTranched;
 
-        // Since Solidity lacks a syntax for copying arrays of structs directly from memory to storage, a manual
-        // approach is necessary. See https://github.com/ethereum/solidity/issues/12783.
-        uint256 count = tranchesWithPercentages.length;
-        for (uint256 i = 0; i < count; ++i) {
+        // Calculate the total percentage of the tranches and save them in the contract storage.
+        uint64 totalPercentage;
```
andreivladbrg, line 55: let's say contract state, as `TOTAL_PERCENTAGE` is immutable and it is a part of the bytecode

andreivladbrg (approved):

src/SablierV2MerkleLT.sol
smol-ninja: https://github.com/sablier-labs/v2-periphery/pull/355/commits/013c5dee0a7aeb9611182b20f9283cc94f4e5daf

PR #357, opened 2024-06-20
feat: function `isPercentagesSum100`

smol-ninja:
### Changelog
- Closes #356 
- Introduces the following view function:
```solidity
/// @notice Verifies if the sum of percentages in `tranches` equals 100% , i.e. 1e18.
/// @dev Reverts if the sum of percentages overflows.
/// @param tranches The tranches with their respective unlock percentages.
/// @return result True if the sum of percentages equals 100%, otherwise false.
function isPercentagesSum100(MerkleLT.TrancheWithPercentage[] calldata tranches)
    external
    pure
    returns (bool result);
```

PR #358, opened 2024-06-20
perf: declare a count for tranches length

andreivladbrg:
Depends on https://github.com/sablier-labs/v2-periphery/pull/357

### Changelog

Declare a `count` variable to optimize loading it into memory on every iteration. It was necessary to add a new scope block to prevent the stack too deep error.


smol-ninja (approved):
LGTM except the build failed.

PaulRBerg (approved):
good, will merge later once the other PR is merged in

PR #329, opened 2024-04-12
docs: roll 1.2.0

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/v2-periphery/issues/306
- Closes https://github.com/sablier-labs/v2-periphery/issues/352

### Subtasks
- [x] https://github.com/sablier-labs/v2-periphery/pull/355
- [x] https://github.com/sablier-labs/v2-periphery/discussions/311#discussioncomment-9086475

PR #360, opened 2024-07-02
build: npm ignore file in test utils

andreivladbrg:
Similar to https://github.com/sablier-labs/v2-core/pull/936


andreivladbrg:
forgot that it is included in https://github.com/sablier-labs/v2-periphery/pull/329

PR #366, opened 2024-07-07
ci: use `fromJSON` to convert string into digits

smol-ninja:
Fixes similar issue as https://github.com/sablier-labs/v2-core/issues/965 in v2-periphery.

Before changes: https://github.com/sablier-labs/v2-periphery/actions/runs/9828216614
After changes: https://github.com/sablier-labs/v2-periphery/actions/runs/9828236839


PaulRBerg (approved):
LGTM

PR #371, opened 2024-07-18
Rename `RPC_URL` and add more endpoints

andreivladbrg:
Similar to https://github.com/sablier-labs/v2-core/pull/973


CONTRIBUTING.md
```diff
@@ -70,7 +70,7 @@ To make CI work in your pull request, ensure that the necessary environment vari
 repository's secrets. Please add the following variables in your GitHub Secrets:
 
 - API_KEY_INFURA
```
smol-ninja, line 72: `API_KEY_INFURA` is no longer required.

.env.example
```diff
@@ -2,4 +2,4 @@ export API_KEY_INFURA="YOUR_API_KEY_INFURA"
 export EOA="YOUR_EOA_ADDRESS"
```
smol-ninja, line 2: `export API_KEY_INFURA="YOUR_API_KEY_INFURA"` is no longer required.

smol-ninja (dismissed):
LGTM except the comments below.

CONTRIBUTING.md
andreivladbrg: correct

.env.example
andreivladbrg: correct

smol-ninja (approved):

PR #374, opened 2024-07-19
chore: refactor CreateMerkle scripts to deploy merkle lockup instances

smol-ninja:
The objective of  having `CreateMerkleLL` and `CreateMerkleLT` scripts is to deploy instances of merkle lockup and verify them on-chain so that etherscan can automatically verify other lockups deployed by the users.

However, its time consuming to construct the parameters of `run` command. This PR makes it easier to achieve the same objective by setting all parameters inside the script itself.

```shell
FOUNDRY_PROFILE=optimized \
forge script script/CreateMerkleLL.s.sol \
--broadcast --etherscan-api-key $ETHERSCAN_API_KEY \
--rpc-url mainnet --sig "run()" \
--verify -vvv
```

### DIFF
--- a/script/CreateMerkleLL.s.sol
+++ b/script/CreateMerkleLL.s.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity >=0.8.22 <0.9.0;
 
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { ISablierV2LockupLinear } from "@sablier/v2-core/src/interfaces/ISablierV2LockupLinear.sol";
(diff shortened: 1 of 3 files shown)


andreivladbrg (changes requested):
What if we merge these two? Or add a new one `CreateMerkleCampaigns` to deploy both `LL` and `LT` in a single run. wdyt?

smol-ninja:
Etherscan verification is not totally reliable so I'd keep them separate. If combined into a single run and one of the verifications fails, it would add a bit of a work to re-run one script while skipping the other.

andreivladbrg:
Got it, thanks for explaining

andreivladbrg (approved):

PR #375, opened 2024-07-22
Polish tests

andreivladbrg:
Depends on https://github.com/sablier-labs/v2-periphery/pull/374

### Changes
- test: deploy Defaults in Base_Test
- test: remove unneeded assets function in Defaults


test/fork/Fork.t.sol
```diff
@@ -41,10 +41,6 @@ abstract contract Fork_Test is Base_Test, V2CoreFuzzers {
         // Load the external dependencies.
         loadDependencies();
 
-        // Deploy the defaults contract and allow it to access cheatcodes.
-        defaults = new Defaults({ users_: users, asset_: FORK_ASSET });
```
smol-ninja, line 45: Though I will wait for the tests to complete, I think fork test would fail because it uses `FORK_ASSET` and not `DAI`.

andreivladbrg: It doesn't (see ci), because we pass the fork assets where we need to test it, e.g.:

https://github.com/sablier-labs/v2-periphery/blob/91ca871fc24cd6c6c614d9ec2876f421ff55d4ea/test/fork/merkle-lockup/MerkleLL.t.sol#L102

smol-ninja: You're right.

smol-ninja (approved):

PR #384, opened 2024-08-14
style: run forge fmt

andreivladbrg:
Created this PR so that the CI won't fail on each scheduled run.


smol-ninja (dismissed):
This forge formatter has been a pain lately.

andreivladbrg:
> This forge formatter has been a pain lately

Indeed

PR #387, opened 2024-12-23
ci: run cron jobs on Sunday evening

smol-ninja:


PaulRBerg:
Can we do it on Saturday at 1am?

So that there are 26 hours between this cron job and the one in V2 Core.

Many RPCs have 24-hour billing cycles.

smol-ninja:
Done. Also changed time for `ci-fork` to make 1 AM reserved for periphery only.

.github/workflows/ci-fork.yml
```diff
@@ -2,7 +2,7 @@ name: "CI Fork and Util tests"
 
 on:
     schedule:
-      - cron: "0 3 * * 1,3,5" # at 3:00 AM UTC on Monday, Wednesday and Friday
+      - cron: "0 1 * * 1,3,5" # at 1:00 AM UTC on Monday, Wednesday and Friday
```
PaulRBerg, line 5: Should we run this only once per week?

Since V2 Periphery is getting deprecated.

smol-ninja: Yeah sure but shouldn't we do that after the next release of Lockup? Until then, its the one with the most usage in production.

PaulRBerg: Good point

PaulRBerg (approved):
LGTM now, merging

